PR #2, opened 2018-09-12
Add proposal folders to keep raisd proposals from community

steven-zou:
**Add proposal folders to keep raisd proposals from community and also update the README to reflect the change**


steven-zou:
Pass review for this folder creating pr

PR #9, opened 2018-09-26
Add chart signature verification proposal

steven-zou:
**Add chart signature verification proposal**
- add _content folder for keeping the images used in the proposal


lweitzman:
Attached images and those mockups can be seen here: https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/340495007_v1-8_-_Helm_Charts

lweitzman:
image001
image002
image003
image004
image005
image006

OrlinVasilev:
@steven-zou is that still valid or we can close it?

wy65701436:
@steven-zou we can close it since the chartmuseum was planned to be deprecated.

Vad1mo:
I think we should maybe support https://github.com/sigstore/helm-sigstore instead as habor does already support sigstore

OrlinVasilev:
closing as we are deprecating chart museum

PR #12, opened 2018-10-09
Add refactor admin server proposal

stonezdj:
Add proposal for refactor admin server


reasonerjt:
Just curious will we have more details in the data structure of configuration items?  Or it falls out of the scope of the proposal?

proposals/new/refactor_admin.md
```diff
@@ -0,0 +1,214 @@
...
+ | redis_password |                | x  |
+ | redis_db_index |                | x  |
+ | registry_storage_provider_config |                | x  |
+ | registry_custom_ca_bundle |                |  x |
+
+3. Remove adminserver related build script, docker file and code.
+
+4. Refactor configuration item management, provide a unified type conversion, validation, default value setting, read, write.
+
+The structure to store metadata of configure items:
```
steven-zou: Seems missing new config options onboarding flow and corresponding interfaces

proposals/new/refactor_admin.md
```diff
@@ -0,0 +1,214 @@
...
+
+The interface for accessing configurations
+
+```go
+// ConfigClient used to retrieve configuration
+type ConfigClient interface {
+	GetSettingByGroup(groupName string) []models.ConfigEntry
+	UpdateConfig(cfg map[string]string) error
+	UpdateConfigItem(key string, value string) error
+	GetConfigString(key string) (string, error)
```
steven-zou: Those value getters can be defined in a new `Value` interface. `ConfigClient` just responsible for maintaining the key-value relationship.

proposals/new/refactor_admin.md
stonezdj: Updated

PR #17, opened 2018-10-29
Authentication: add oidc support

silenceshell:
for [#5358](https://github.com/goharbor/harbor/issues/5358)


reasonerjt:
Thanks a LOT for this proposal!
Some initial questions:
1)  So this is essentially OIDC support via DEX right?  Means user needs to setup DEX some where, but not **any** OIDC provider?
2)  In your scenario, how do you associate user with project in Harbor?

silenceshell:
1. dex is just an example of OIDC provider. We use go-oidc package to communicate with OIDC provider.
2. We have an sso service based on a LDAP server, which is also the authentication server for Harbor. We add an connector for the sso service to Dex. When user login, Dex could call the sso service api to get user info, which is the same with the LDAP server. Then the `oidcTokenReqCtxModifier` in harbor could call Dex api to verify the token and get user info by `IDToken.Claim`. Thus the user in OIDC associate with user in Harbor.
The sso service is not essential, we could use LDAP as a Dex connector directly.
harbor

clouderati:
Hi @silenceshell – many thanks for writing this up!

I am also curious about the answers to @reasonerjt's questions. Also wondering how we scope / limit which users can authenticate to Harbor in this scenario. To be more specific, if `dex` is auth'ing against Active Directory and we want users in a specific AD / LDAP group to be able to authenticate, is there a way to filter by group? Is this something done in `dex`, Harbor or both?

How easy is it to set up `dex` + Harbor in this particular scenario? When you start working a PR can you please be certain to set up a document that describes an end-to-end configuration that users can leverage?

Thanks again!

silenceshell:
hi @clouderati 
Dex has filter for user/group(as shown below). Do you mean this?

```yaml
    connectors:
    - type: ldap
      name: OpenLDAP
      id: ldap
      config:
        host: ldap-service:389
        insecureNoSSL: true
        bindDN: cn=admin,dc=example,dc=org
        bindPW: admin
        usernamePrompt: Email Address
        userSearch:
          baseDN: ou=People,dc=example,dc=org
          filter: "(objectClass=person)"
          username: mail
          idAttr: DN
          emailAttr: mail
          nameAttr: cn
        groupSearch:
          baseDN: ou=Groups,dc=example,dc=org
          filter: "(objectClass=groupOfNames)"
          userAttr: DN
          groupAttr: member
          nameAttr: cn
```

I have set up an env for this PR, and will commit the config files of dex/ldap/harbor/example-app (maybe code of harbor also?) and describe more clearly in this PR.

silenceshell:
@reasonerjt @clouderati 
I pushed code to my [repo](https://github.com/silenceshell/harbor/commit/75dc9a93f26a82eaa70f00e1d357ca6483e7c927), it is based on the lastest master (maybe not the newest). Maybe code explains better.

`dex.example.com` is the hostname of dex server, shoule be set according to your env.

Harbor use OIDC authentication as [Kubernetes authentication through dex](https://github.com/dexidp/dex/blob/master/Documentation/kubernetes.md), you could set up by referencing to this doc, it needs `example-app` to generate token.

reasonerjt:
@silenceshell Per discussion we've decided to add OIDC support in our 1.8 release, will you have bandwidth to work together to refine the proposal, and implement the feature?

For current proposal, I have some additional questions:
1)  For E2E scenario, we should clarify the process for user to add or configure OIDC providers?  
2) After configured the OIDC provider, we should also allow some users to login via login form?  such as the local admin user?
3) In your diagram the flow for docker cli relies on `basic auth`, does it mean the users in the dex cannot `docker login`?  That would be a problem.

silenceshell:
@reasonerjt With pleasure to work together and implement this feature :)

1. I think it is not a 'must'. May be we could add some suggestion as [k8s](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)
2. Yeah. OIDC is another method for authentication, local admin user could still login from Harbor web site.
3. Dex could use LDAP as original auth source(and [others](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)), If end user configs  basic auth for Harbor based on the same LDAP service, users in LDAP could still login from docker client. If it is based on database, I am afraid indeed it is a problem.

reasonerjt:
1. For E2E scenario, given how Harbor works now, I hope we can provide mechanism for admin to configure OIDC provider.
3. I don't think we can rely on user to configure the same identity source in Harbor and Dex or other OIDC provider, but the user experience may be bad as docker CLI may not work like browser to handle the redirections ,etc.  But consistency is more important, more investigation is needed here..

danielzhanghl:
hi, to connect to keycloak, there are other options, in this PR, I reuse most of OAUTH2 code to connect to keycloak,  which is familiar UAA, most to the code are common,  and the changes:
1.add a new configuration option realm, which is keycloak term.
2.the end point/structure of UAA and keycloak are different for search user,/authorization, I separate d these two.
https://github.com/goharbor/harbor/pull/6425


for keyclaok side, it's need to configure service account with list user permission beside a client/realm

for example, below are the harbor configurations:
#The following attributes only need to be set when auth mode is uaa_auth
uaa_endpoint = https://keycloakendpoint/auth
uaa_clientid = harbor-client
uaa_clientsecret = 86e44958-1d2d-4412-aa02-a78c59698b0e
uaa_verify_cert = false
realm = harbor-realm

keycloak settings:

image


image

reasonerjt:
@danielzhanghl I don't think this proposal should cover vendor-specific terms such as "realm", the goal is to use the same code path to provide a consistent flow to integrate with any OIDC provider.

reasonerjt:
Per discussion with @silenceshell I'm closing this PR and create a new one to cover the areas I mentioned in my comment.

ese:
is there anybody working on this? I tried to follow the clue through some issues and seems that the state is the work done in the @silenceshell 	fork https://github.com/silenceshell/harbor/commit/75dc9a93f26a82eaa70f00e1d357ca6483e7c927

steven-zou:
@ese 

The revised ODIC supporting proposal is drafting and will be submitted later for review.

PR #19, opened 2018-11-05
add proposal of support metadata

wy65701436:


steven-zou:
@wy65701436 

DCO missed!

proposals/new/metadata.md
```diff
@@ -0,0 +1,107 @@
...
+Data model
+type metadata struct {
+	ID uint64 `json:"id"`
+	SourceID int64 `json:"source_id"`
+	Name string `json:"name"`
+	Value string `json:"value"`
+	Description string	`json:"description"`
+	CreationTime time.Time    `json:"creation_time"`
+	UpdateTime   time.Time    `json:"update_time"`
+}
```
cd1989:
```
type metadata struct {
	ID uint64 `json:"id"`
	SourceID int64 `json:"source_id"`
	Name string `json:"name"`
	Value string `json:"value"`
	Description string	`json:"description"`
	CreationTime time.Time    `json:"creation_time"`
	UpdateTime   time.Time    `json:"update_time"`
}
```
Would be easier to read.

proposals/new/metadata.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Solution
+CI/CD system uploads the images/helms together with build/code information into Harbor, which enables the capability of developer to view these in registry, and make it easy to trace code commit and build pipeline message. 
+This proposal describes three parts, that are how to publish metadata into Harbor, how to present metadata in Harbor portal and how to use metadata in Harbor.
+
+## Proposal 
+1, How to publish metadata into Harbor — Harbor provides API to handle CRUD of metadata, to enable fully tracebable docker builds and charts.
+
+API
+1, /api/{source}/{source_id}/metadata   POST
```
cd1989: What does `source` mean here ?

reasonerjt:
Current proposal defines metadata of an image as another resource.  
IMO, we need to find a better way to ensure the integrity between image and metadata.
For example, a user may push an image to overwrite an existing tag via docker CLI, and the metadata will no longer reflect the actual status of the new image.
Associating the metadata to the `sha256` may mitigate such issue.

Would it be better if we can investigate the possibility to extend the manifest of an image, say by `LABEL`, such that the metadata is baked into the image, and there won't be inconsistency caused by our API?

paroque28:
Any updates regarding this proposal?

Vad1mo:
Having an API is nice, but from an adoption perspective, it will make it challenging for users to integrate that into into their pipelines. A CLI can facilitate that. 

There is another solution to how the CI/CD pipeline results can be displayed in Harbor and it already exists but harbor doesn't display it yet.
 
https://github.com/opencontainers/image-spec/blob/main/annotations.md this is based on some efforts in the past by http://label-schema.org/rc1/ 

So if Harbor could display some of these annotations, this would help create a link and use it also for other cases.

Vad1mo:
@wy65701436, I can make an additional proposal focussing on only that!

wmgroot:
Did this ever result in a new proposal and implementation? My team is investigating if we can use this feature with the latest version of Harbor to track which git repositories an image was built from as metadata.

Vad1mo:
Annotations is a good start @wmgroot also in 2.5 you have the option to use the `accessory` functionality. Which is essentially an attachment (as on oci-image) to a image. It is used to link co-sign signatures with images.

PR #22, opened 2018-11-22
Add Proposal of Refactor prepare_script

ninjadq:
propose to refactor `prepare` script


steven-zou:
@ninjadq 

Please squash your commits and fix the DCO issue.

proposals/new/propose_to_refactor_prepare_script.md
```diff
@@ -0,0 +1,371 @@
...
+   
+   ```
+
+3. The `/data` dir is hardcoded in template file, this can also be solved easily in replacing render engine. Files including hardcoded `/data` are `docker-compose.yml`, `harbor.cfg`, `.travis.yml`, `prepare`, `core/systeminfo/systeminfo.go`, `jobservice/generateCerts.sh`,
+`tests/docker-compose.test.yml`, `tests/resources/Harbor-Util.robot`, `tests/robot-cases/Group2-Longevity/Longevity.robot`,
+`tests/testprepare.sh`, `tests/travis/api_common_install.sh`, `tests/travis/ut_install.sh`
+
+
+4. Current config file is `harbor.cfg` which is configparser-style format. In order to describe more complex content, We need replace it using YAML format like `harbor.yml` . current config file may look like this.
+```yaml
```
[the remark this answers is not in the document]
ninjadq: Agree with you, the final config file should like this. But the first step is compatible with the previous configParser format.

reasonerjt:
Two other goals we should consider to achieve in this refactor:
1)  The default data volume `/data` should be configurable.  Particularly, the `secretkey` should be moved away from the data volume.  @ywk253100 knows more background.
2)  The `sudo` should be removed from entrypoints of the containers in Harbor.

reasonerjt:
We also need a solution for migration from old `harbor.cfg` to `harbor.yaml`
I think that's do-able following the way we did, refer to:
https://github.com/goharbor/harbor/tree/master/tools/migration/cfg

proposals/new/propose_to_refactor_prepare_script.md
```diff
@@ -0,0 +1,371 @@
...
+| clair_db_port           | number  |
+| clair_updaters_interval | number  |
+| customize_crt           | boolean |
+| email_insecure          | boolean |
+| ldap_verify_cert        | boolean |
+| self_registration       | boolean |
+| uaa_verify_cert         | boolean |
+
+
+5. Packaging all files and its denpendencies into a container.  All these codes and changes is related to python, so we should using official python images as the base image in Dockerfile. But there is a issue that some action need super user to operate. like change the owner and group of an file, create files on host system. I have no proper method to handle this. Maybe mount the directory on host specified in config file to container can solve last question.
```
steven-zou: Suggest adding an upgrading section to talk about how to upgrade config from old version to new version.

proposals/new/propose_to_refactor_prepare_script.md
```diff
@@ -0,0 +1,371 @@
...
+   ```
+
+3. The `/data` dir is hardcoded in template file, this can also be solved easily in replacing render engine. Files including hardcoded `/data` are `docker-compose.yml`, `harbor.cfg`, `.travis.yml`, `prepare`, `core/systeminfo/systeminfo.go`, `jobservice/generateCerts.sh`,
+`tests/docker-compose.test.yml`, `tests/resources/Harbor-Util.robot`, `tests/robot-cases/Group2-Longevity/Longevity.robot`,
+`tests/testprepare.sh`, `tests/travis/api_common_install.sh`, `tests/travis/ut_install.sh`
+
+
+4. Current config file is `harbor.cfg` which is configparser-style format. In order to describe more complex content, We need replace it using YAML format like `harbor.yml` . current config file may look like this.
+```yaml
+## Configuration file of Harbor
```
steven-zou: Please append the property table copied from daojun's PR to let reviewers know what settings will be removed from `harbor.yaml`.

nlowe:
I know the existing `prepare` script is written in python. Have we considered re-writing it in `go` instead, and using go templates? It drops the dependency on Python at runtime and brings the utility closer in-line with the majority of the codebase.

As far as I can tell, python is only used in some test scripts and the CLI tool which is deprecated.

ninjadq:
> I know the existing `prepare` script is written in python. Have we considered re-writing it in `go` instead, and using go templates? It drops the dependency on Python at runtime and brings the utility closer in-line with the majority of the codebase.
> 
> As far as I can tell, python is only used in some test scripts and the CLI tool which is deprecated.

I think the effort of rewriting it into `go` and `go template` is large then refactor in python because we can reuse a lot of current codes and these codes are verified by time. In addition, we wanna put all the dependencies of `prepare` into a container, so that user does not care about the python dependencies anymore. `go` is excellent but from my knowledge, python is good at this kind scripts.

clouderati:
I have some relatively broad thoughts on the topic of having to run a `prepare` script in the first place. I'm going to parse those thoughts and eventually submit a fleshed out proposal.

With that said, though, I do agree with @nlowe that having multiple components in disparate languages is something we should eventually aim to address. We should pick one _lingua franca_ and stick with it as much as possible (within reason).

steven-zou:
Per lazy consensus rule, merge proposal to main.

PR #28, opened 2018-12-11
add project quota proposal

danielzhanghl:
Signed-off-by: daniel zhang <danielzhanghl@gmail.com>

### DIFF
--- a/proposals/new/quota-propsoal.md
+++ b/proposals/new/quota-propsoal.md
@@ -1,0 +1,44 @@
+# Proposal:`<Daniel Zhang>`
+
+Author: `<Daniel Zhang/ danielzhanghl>`
+
+Discussion: `https://github.com/goharbor/harbor/issues/6271`
+
+## Abstract
+
+add docker repo quota to project.
+
+## Background
+
+[harbor has no restrictions on repository and tag quotas.For example: how many repositories can be created per project and how many tags can be created per repository. And This quota should be set by the user. The problem it caused.When a repository has many tags, the query is slow.]
+
+## Proposal
+
+1.project quota could be set/modified when create/modify project, it is used to limit the images that user could push, for example, 1024M(it's default value).
+
+2. image size is queried from image manifest file: size of all layers and manifest file size.
+when user upload a layer or manifest, project usage will be checked with quota, if it's going to exceed, the load will be blocked, otherwise, the project usage will be counted.
+
+3. when image is pushed, the usage will be re-calculated by reading all the images manifest in the project, therefore, there maybe "usage" great than "quota", but it's not the real storage situation, as docker client did not upload the layer but only refer it in the images manifest file. and the next upload attempt will be blocked as the usage is great than quota.
+
+4. if user push two images, with part of or all the layers are same, the total size used is counted twice though the layers are shared.
+
+
+## Non-Goals
+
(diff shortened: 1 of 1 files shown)


proposals/new/quota-propsoal.md
```diff
@@ -0,0 +1,44 @@
...
+
+add docker repo quota to project.
+
+## Background
+
+[harbor has no restrictions on repository and tag quotas.For example: how many repositories can be created per project and how many tags can be created per repository. And This quota should be set by the user. The problem it caused.When a repository has many tags, the query is slow.]
+
+## Proposal
+
+1.project quota could be set/modified when create/modify project, it is used to limit the images that user could push, for example, 1024M(it's default value).
```
cd1989: I think there is inconsistency between the _background_ and the _proposal_, where the former talks about repo/tag `number`, while the latter talks about `image size`.

proposals/new/quota-propsoal.md
```diff
@@ -0,0 +1,44 @@
...
+when user upload a layer or manifest, project usage will be checked with quota, if it's going to exceed, the load will be blocked, otherwise, the project usage will be counted.
+
+3. when image is pushed, the usage will be re-calculated by reading all the images manifest in the project, therefore, there maybe "usage" great than "quota", but it's not the real storage situation, as docker client did not upload the layer but only refer it in the images manifest file. and the next upload attempt will be blocked as the usage is great than quota.
+
+4. if user push two images, with part of or all the layers are same, the total size used is counted twice though the layers are shared.
+
+
+## Non-Goals
+
+1. the storage of docker repo are shared with the entire repositroy, there is no exact way to control to the usage for a specific project, this propsal is to limit the total size declared in one project, not the real storage.
```
cd1989: I think the `declared total size` doesn't make much sense, as it doesn't reflect the real resource usage. Implement project quota on this doesn't seem a good way. For example, if user retags images in the same project, it will still be counted in the total usage, this would confuse users.

I suggest to either put constraints on repo/tag numbers, or constraints on real storage usage.

reasonerjt:
In addition to Storage usage, there may be different kinds of Quota.
I would suggest firstly we define what Quota controls and provide a extensible "Quota" management framework.
@steven-zou, I believe it is required that there should be technical designs in proposal?

proposals/new/quota-propsoal.md
```diff
@@ -0,0 +1,44 @@
...
+## Proposal
+
+1.project quota could be set/modified when create/modify project, it is used to limit the images that user could push, for example, 1024M(it's default value).
+
+2. image size is queried from image manifest file: size of all layers and manifest file size.
+when user upload a layer or manifest, project usage will be checked with quota, if it's going to exceed, the load will be blocked, otherwise, the project usage will be counted.
+
+3. when image is pushed, the usage will be re-calculated by reading all the images manifest in the project, therefore, there maybe "usage" great than "quota", but it's not the real storage situation, as docker client did not upload the layer but only refer it in the images manifest file. and the next upload attempt will be blocked as the usage is great than quota.
+
+4. if user push two images, with part of or all the layers are same, the total size used is counted twice though the layers are shared.
```
reasonerjt: This may cause issue when user pushes image to Harbor via a CI system.

proposals/new/quota-propsoal.md
danielzhanghl: the real storage usage is hard to limit as it's shared by images, if the "declared total size" is not enough, could add others, for example, tags number...

reasonerjt:
Closing due to inactivity, #81 will cover such use cases.

PR #33, opened 2018-12-27
add new sharinging  slides from kubecon seattle conf

steven-zou:


steven-zou:
Just adding new sharing slides, pass the approval stage.

PR #35, opened 2019-01-04
Add proposal of robot account

wy65701436:


proposals/new/robot.md
```diff
@@ -0,0 +1,385 @@
...
+
+````
+The robot account cannot login to harbor portal.
+
+No code change as all the robot accounts are stored in harbor_robots instead of harbor_user.
+````
+
+### Docker login
+
+To distinguish the robot account from user, it will add a predefined prefix "harborobots" to the name of robot, like "harborobots_example1".
```
nlowe: While this would definitely make the auth code easier to implement, is there a way we can do this without a user prefix or at least a shorter one?

What if we just make the token unique, and not give a "robot" a name? Every team that has a robot account would use something like `ci` or `harborrobot` (configurable at the server level by the server admin) and provide their team specific token as the password? 

I think this makes supporting segregated permissions for multiple teams **much** simpler, since teams don't have to prefix/suffix their bot names (i.e. `jennkins-team-a`, `jenkins-team-b`, `jenkins-infra-team`, etc. etc.)

proposals/new/robot.md
```diff
@@ -0,0 +1,385 @@
...
+
+1, with user/pwd
```
nlowe: It's not explicitly spelled out but I think these two examples would be two good things to include in the UI upon creation of the robot account.

proposals/new/robot.md
```diff
@@ -0,0 +1,317 @@
...
+
+201 ok
+{
+  name: "test1",
+  token: "rKgjKEMpMEK23zqejkWn5GIVvgJps1vKACTa6tnGXXyOlOTsXFESccDvgaJx047q"
+}
+
+````
+
+2, Disable a robot account
```
nlowe: Should this be **Delete** a robot account?

proposals/new/robot.md
```diff
@@ -0,0 +1,317 @@
...
+3, View a/all robot account
+
+````
+
+GET /api/project/${id}/robots
+GET /api/project/${id}/robots/${id}
+
+````
+
+It's actually to issue an new token bases on the updates in the background.
```
nlowe: Can you clarify what this means?

wy65701436: the notes should be removed with PUT api.

proposals/new/robot.md
wy65701436: it actually deletes a robot account but in UI, it's called disable an account.

nlowe: Disabling implies it can be re-enabled without needing to be re-created. One of these needs corrected, does it actually delete the account or disable the account?

proposals/new/robot.md
wy65701436: actually, this proposal is focused on the docker login process, just mention that UI doesn't support robot authn.

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+	if !ok {
+		return false
+	}
+	if !strings.HasPrefix("username", "harborobots") {
+		return false
+	}
+	log.Debug("got user information via basic auth")
+
+    // Decode the token to validate it in the DB.
+	user, err := auth.LoginRobot(models.AuthModel{
```
reasonerjt: I recall we decided we should not do this, please update the proposal accordingly.

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+type ResourceActions struct {
+	Name    string   `json:"name"`
+	Actions []string `json:"actions"`
+}
+
+```
+
+##### Scpoe -- Implement a new security context for robot account
+
+In the implementation of robot context, the func of HasReadPerm and HasWritePerm could validate the token scope, and block all the request to harbor with forbidden.
```
reasonerjt: 1. I don't think it needs `project manager` in the robot context
2. If we do not implement `RobotLogin`, the token should be validated when creating the Security Context.

proposals/new/robot.md
nlowe: So are we only supporting the creation of robot accounts via the API?

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+
+````
+The robot account cannot login to harbor portal.
+
+No code change as all the robot accounts are stored in harbor_robots instead of harbor_user.
+````
+
+### Docker login
+
+To distinguish the robot account from user, it will add a predefined prefix "harborobots" to the name of robot, like "harborobots_example1".
```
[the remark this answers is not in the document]
wy65701436: use the robot$ as the prefix.

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+    {"name":"project/repo", "actions":["read"]},
+    {"name":"project/label", "actions":["write"]}
+  ]
+}
+
+````
+
+````
+
+201 ok
```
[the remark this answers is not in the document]
wy65701436: did it in the PR, will update the document here.

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+
+```yaml
+
+CREATE TABLE robot (
+ id SERIAL PRIMARY KEY NOT NULL,
+ name varchar(255),
+ /*
+  The maximum length of token is 7k
+ */
+ token varchar(7168),
```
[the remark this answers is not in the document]
wy65701436: yes, all of API are not depends on this column, it could be removed anytime.

proposals/new/robot.md
wy65701436: yes, will update the code segment.

proposals/new/robot.md
```diff
@@ -0,0 +1,326 @@
...
+##### Config items
+
+| Item               | Value          | Level  | Note                  |
+| ------------------ | -------------- | ------ | --------------------  |
+| Robot_Token_Expire | 30             | User   | Hard code for 1 year  |
+| Robot_Token_Key    | /etc/robot_key | System | Share the key of core |
+
+##### Model
+
+```go
```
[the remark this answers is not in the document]
wy65701436: I will create a new token service without modifying the existing one by the end of February.

proposals/new/robot.md
wy65701436: updated the code segment

steven-zou:
@nlowe @cd1989

If no further comments, let's `+1` to merge it into the master.

steven-zou:
@wy65701436 

DCO missing.

steven-zou:
> While I still think we should seriously consider [using a server-wide configured username](https://github.com/goharbor/community/pull/35#discussion_r245550889) instead of prefixes for `docker login`, I don't think that should block this PR and can be addressed later. Overall, LGTM.

@nlowe 

I do not see the significant meaningful benefits of replacing the fix robot user name `robot$` with the configurable username. Let's see if we can get more same requirements from the community after user trying it.

proposals/new/robot.md
```diff
@@ -0,0 +1,321 @@
...
+
+````
+
+POST /api/project/${pid}/robots
+Data: 
+{
+  name: "test1",
+  desc: "test1_desc",
+  access: [
+    {"name":"project/repo", "actions":["pull"]},
```
zhoumeina: If the access will be a list , should it be "actions"? and using array? I think it should be "action":"pull"

reasonerjt: 1.  For extensibility, we'd better make the access include the type of resource.  Would it make sense to add a field `type` in the item of access?  Or name it like `"resource":"repository:project/repo"`. 
2. If the `actions` is a list, in this example, why `pull` and `push` are in two items.
3. What if the repo name point to a repo in another project?  What if there are images pushed after the robot is created?

wy65701436: the access will align with policy defined in the rbac, things like:

{"resource":"/projects/1/repo", "actions":"pull"},
{"resource":"/projects/1/repo", "actions":"push"}

I will update the doc accordingly.

steven-zou:
@nlowe @cd1989 

Can we merge this PR now?

proposals/new/robot.md
wy65701436: doc updated.

proposals/new/robot.md
clouderati: I think we need to support both creation and removal via UI. Just thinking through the workflow here, the individual(s) responsible for leveraging the robot accounts are going to want to quickly create the user, then copy the credentials into some CI system like Jenkins.

Requiring a `curl` command to create a robot user is a bit tedious IMHO.

clouderati:
Our DCO robot friend is unhappy. Do you mind both squashing commits and signing before we merge?

(note that there are a few questions / outstanding comments to address anyway)

PR #37, opened 2019-01-08
Proposal for health check API

ywk253100:
Signed-off-by: Wenkai Yin <yinw@vmware.com>


proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+    "enabled": true,
+    "status": "healthy"
+  },
+  "clair": {
+    "enabled": false
+  },
+  ...
+}
+```
+Response(unhealthy), the status is unhealthy as long as the status of one of the components is unhealthy:
```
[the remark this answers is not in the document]
ywk253100: The components are closely associated with others, if one of them is unhealthy, the function may not be used: when the `Clair` doesn't work well, the images that enable `Prevent vulnerable images from running` cannot be pulled either. So if users don't care about the `Clair`, maybe they should disable it when installing Harbor?

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+```
+200 OK
+{
+  "stauts": "healthy",
+  "core": {
+    "enabled": true,
+    "status": "healthy"
+  },
+  "clair": {
+    "enabled": false
```
steven-zou: Do we need to return the system which is not installed/enabled? IMO, we just need to return the healthy status of the installed ones.

What's the use case of the return `enabled` flag in the health API?

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+  },
+  ...
+}
+```
+Response(unhealthy), the status is unhealthy as long as the status of one of the components is unhealthy:
+```
+503 Service Unavailable
```
steven-zou: 503 is correct? Per my understanding, 503 means the `/api/health` API calling is failed/not accessible.

If the heath API does work, it should always return the overall health status info with 200 code.

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+{
+  "stauts": "unhealthy",
+  "core": {
+    "enabled": true,
+    "status": "unhealthy",
+    "detail": "here is the error message"
```
steven-zou: An optional error may be more suitable than 'details' because it's more meaningful.

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+  },
+  ...
+}
+```
+
+The components that health check API exposes:
+
+Component | Check health | How to check | Comment 
+------------|------------|------------|------------
+ | Proxy | Yes | HTTP `GET /` | Not check int Harbor chart when exposing service via ingress or load balancer, so need a way to configure it
```
steven-zou: I don't think we need to check proxy because your api is behind proxy. If the proxy is down, your API will be going down.

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,90 @@
...
+ | Core | Yes | Internal method call |
```
steven-zou: If no subcomponents to check, we can ignore this core check like the above proxy.

ywk253100:
The proposal is updated according to the comments and review.

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,100 @@
...
+## Background
+
+The stability and availability are essential for the modern applications. The awareness of health status of an application is the precondition to make corresponding decisions when encountering software failure.  
+
+The health check API introduced by this proposal can be used to integrated with other monitor systems to help operators manage the Harbor service easier.
+
+## Implementation
+Using pull mode, health check API collects status of every components and returns them to the client.  
+
+We'll implement a `HealthChecker` for every components, and the components can determine how to check its status. The `HealthChecker` can collect the status and store it in the memory periodically to reduce the latency.
```
nlowe: I'm assuming this is intended to be an interface? Can you include an example of what you think it would look like?

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,100 @@
...
+```
+GET /api/health
+```
+
+Response(healthy):
+```
+200 OK
+{
+  "status": "healthy",
+  "components": [
```
nlowe: Is this always a flat-map of components? Do we care about showing things that are unhealthy because one or more of their dependencies are unhealthy?

nlowe: If we don't care about showing the health relationships between dependent components, what about making this a simple map instead:

```json
{
  "status": "healthy",
  "components": {
    "core": "healthy",
    "clair": "healthy",
    "...": "...",
  }
}
```

ywk253100: We need to return the error message for the component to explain why it is unhealthy.

proposals/new/health-check-api.md
ywk253100: Yes, it is an interface and I have added it on the proposal

proposals/new/health-check-api.md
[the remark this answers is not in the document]
cd1989: Clair is just one example, what I want to express is: can I find out whether a function of Harbor is working well, not a global status of Harbor (healthy, unhealthy). Because Harbor includes lots of functions, registry, clair, notary, helm chart, ... unhealthy of some parts of them should not block my using of the normal functions.

For example, if my CI/CD is simple, I don't care about clair but just I want my image pushed as long as the registry is working well. But this doesn't mean I don't need clair in other cases. Or maybe notary is a better example.

Or put in another way, I'd suggest health check focus on functions, not on components, components status can be details info attached in the result.

proposals/new/health-check-api.md
```diff
@@ -0,0 +1,109 @@
...
+
+The configurations need to be added in `core`:
+
+Name | Comment 
+------------|------------
+ | PORTAL_URL | The URL of portal 
+ | JOBSERVICE_URL | The URL of jobservice 
+ | REGISTRY_URL | The URL of registry 
+ | REGISTRYCTL_URL | The URL of registryctl 
+ | CHARTMUSEUM_ENABLED | Enable chartmuseum or not
```
steven-zou: Suggest to use a separate section to describe how to get the installed components and what configurations needed.

proposals/new/health-check-api.md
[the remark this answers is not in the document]
ywk253100: Notary is the same case, the images cannot be pushed/pulled if Notary doesn't work well in some cases even though the Registry works as expected. This is because that the components are closely associated with others. One function can work well depends on not only a single component.

For the suggestion that the API should focus on functions, it's hard to define what is core function as core function means different features for different users.

PR #26, opened 2018-12-01
Pull images from Docker Hub to Harbor

cd1989:
Signed-off-by: 陈德 <chende@caicloud.io>

PR #21, opened 2018-11-14
Harbor support images replication to public cloud registry

yuanshuhan:
Harbor support images replication to public cloud registry

PR #40, opened 2019-01-10
Add proposal for replication NG(next generation)

steven-zou:
Add proposal for replication NG(next generation)


proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+However, the current solution provides very limited capabilities. Many replication scenarios cannot be covered. It should be improved to extend the usage scope and scenarios. The goal of this proposal is to propose a enhanced replication solution (named as replication NG solution) to achieve the capabilities:
+
+* Replicate both image and chart
+* Replicate content from Harbor to other non Harbor registries
+* Replicate content from other registries to Harbor
+* Easily extended to onboard the replication between Harbor and a new non Harbor registry
+
+## Proposal
+
+### Use Cases
```
nlowe: Nit, odd unicode character before the `U` in `Use Cases`

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+
+The replication NG should cover the following use cases:
+
+* User can manage and monitor both Harbor and non-Harbor registry endpoints
```
nlowe: What does monitoring an endpoint mean?

proposals/Content-Replication-ng.md
nlowe: What's odd is this ***didn't*** break the anchor link to this later on in the document!

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+### Overall Architecture
+
+Here is the over all architecture for the replication NG service.
+
+![replication NG architecture](images/replica_ng_arch.png)
+
+#### Restful API
+
+The replication restful APIs are supported by three parts:
+**WARNING: New API routes breaks the compatibilities**
```
nlowe: Will this cause the version of Harbor itself to be bumped to 2.x? If so, is there any other breaking changes that should be batched alongside this?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+  ```
+
+  * **NG API route definitions**:
+
+  ```go
+  /**
+  * 1. As the managed registries may be sources or targets in replication NG, the `targets` term in the API routes of registry management is not suitable anymore. Use `registries` to instead.
+  * 2. As the health checking is handled by the backend monitor, ping can be moved out.
+  * 3. API of listing policies by target is also moved out.
+  * 4. Left an open question here: is it still necessary to limit the ":id" with only numbers?
```
nlowe: If an `id` isn't `[0-9]+`, what else did you have in mind?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+
+  * **NG API route definitions**
+
+  ```go
+  /**
+  * 1. Add `replication` as a API group context to indicate those APIs handle replication related things.
+  * 2. Each replication running based on the policy will be recorded as an `Execution`.
+  * 3. Move the DELETE replication execution out.
+  */
+  POST "/api/replication/executions"                 // Start/trigger a replication execution
```
nlowe: How do we determine what execution to trigger?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    // The key of the access account, for OAuth token, it can be empty
+    AccessKey string
+    // The secret or password for the key
+    AccessSecret string
+}
+
+// Namespace represents the full path of resource isolation unit;
+// if the namespace has hierarchical structure, e.g organization->team,
+// it should be converted to organization.team
+type Namespace string
```
nlowe: Do we need a method to normalize namespaces to the format outlined in the documentation comment?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+// ResourceType represents the type of the resource.
+type ResourceType string
+
+// Resource represents the general replicating content
+type Resource struct{
+    Type         ResourceType           `json:"resource_type"`
+    URI          string                 `json:"uri"`
+    Metadata     *Metadata              `json:"metadata"`
+    Registry     *Registry              `json:"registry"`
+    ExtendedInfo map[string]interface{} `json:"extended_info"`
```
nlowe: What do you see this being used for?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+type Policy struct {
+    // metadata
+    ID             string     `json:"id"`
+    Name           string     `json:"name"`
+    Description    string     `json:"description"`
+    Creator        string     `json:"creator"`
+    CreationTime   time.Time  `json:"creation_time"`
+    UpdateTime     time.Time  `json:"update_time"`
+    // source
+    SrcRegistry    int64      `json:"src_registry"`
```
nlowe: This should probably be named `SrcRegistryID`

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    SrcNamespaces  []string   `json:"src_namespaces"`
+    // destination
+    DestRegistry   int64      `json:"dest_registry"`
```
nlowe: This should probably be named `DestRegistryID`

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    DestNamespaces []string   `json:"dest_namespaces"`
```
nlowe: if I have multiple source namespaces and multiple destination namespaces, how do I know what's being replicated to what?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    // Filters
+    Filters        []*Filter  `json:"filters"`
+    // Triggers
+    Triggers       []*Trigger `json:"triggers"`
+    // Settings
+    Deletion       bool       `json:"deletion"`
```
nlowe: What does this signify?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    // Operations
+    Enabled        bool       `json:"enabled"`
+}
+
+// FilterType represents the type info of the filter.
+type FilterType string
+
+type Filter struct {
+    Type    FilterType  `json:"type"`
+    Value   interface{} `json:"value"`
```
nlowe: Perhaps this is me being ignorant of how filters are currently used, but can you give an example of how the `Value` property would be used?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    // Update the specified policy
+    UpdatePolicy(policy *Policy)error
+    // Remove the specified policy
+    RemovePolicy(ID string)(*Policy, error)
+}
+```
+
+#### Registry Management
+
+The registry endpoints are managed by the registry manager. From registry manager perspective, it only handles the data models of registry endpoint. A daemon monitor will help to continuously check the health of all the managed registry endpoints and update the healthy status in the data model of registry endpoint.
```
nlowe: Do we really need to know the health of remote endpoints at all times? Doesn't it only matter when replication actually happens or when a policy is created / updated?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+```go
+// Indicate the type of registry
+type RegistryType string
+
+// Registry keeps the related info of registry
+// Data required for the secure access way is not contained here.
+// DAO layer is not considered here
+type Registry struct {
+    // UUID of the registry endpoint
+    ID string              `json:"id"`
```
nlowe: Why does this use a UUID for the ID while other models use an `int64`?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+**NG registry data model**:
+
+```go
+// Indicate the type of registry
+type RegistryType string
+
+// Registry keeps the related info of registry
+// Data required for the secure access way is not contained here.
+// DAO layer is not considered here
+type Registry struct {
```
nlowe: Should we add a way to ignore certificate errors (a separate concern for supporting insecure registries, which are served as `http://`)? Or will we require consumers to install certificates into the job service container?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+}
+```
+
+#### Replication Operation
+
+The replication operation module take the responsibility of drive the overall flow we defined in [Process flow](#process-flow).
+
+The `Controller` component provides unified access interfaces to the restful API layers based on the backend components.
+
+The `Replication Flow Controller` accepts the replication call from the `Controller` and start to initialize the flow process. Get the policy from the `Policy Manager` and parse it and inject into the flow. Get the registry data from the `Registry Endpoint Manager` and wrap data into the flow. Base on the above data, retrieve and constructed related registry adapters from the `Adapter Manager`. Retrieve resources (image/chart/all) from the specified source registry and apply filter polices to get the replicating resource candidates.
```
nlowe: So if I have this right, the `Controller` exposes the HTTP API surface, and the `Replication Flow Controller` generates replication jobs?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+
+The `Scheduler` schedule jobs to do the real data transferring of the resource candidates generated by the `Replication Flow Controller` by calling job service APIs. The `Scheduler` should follow the below interface:
+
+```go
+// Schedule tasks to transfer resource data
+type Scheduler interface {
```
nlowe: Doesn't harbor already have a scheduler component? Is this a reference to the existing component? A new one?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    // Schedule tasks
+    Schedule(resources []*Resource) error
```
nlowe: What is the `[]*Resoruce` here? A slice of resources **to** replicate? Alternating pairs of `src` and `dest` Resources?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+}
+
+```
+
+The `Hooks` listens to the job execution status/result via web hook mechanism and update those data to the replication record via `Replication Manager`. The hook route can be:
```
nlowe: Is this notifying the Replication manager of new work to do, or external parties that we're about to do a replication?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+> Start Timestamp: 2019/01/01 12:00:00
+> Duration: 2 minutes
+> Logs:
+> job log 1 ...
+> job logs 2 ...
+
+```go
+// Result summary of the once replication execution.
+type Execution struct{
+    ID         string      `json:"id"`
```
nlowe: What's the format of this ID?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    PolicyID   int64       `json:"policy_id"`
+    Total      int         `json:"total"`
+    Failed     int         `json:"failed"`
+    Succeed    int         `json:"succeed"`
+    InProgress int         `json:"in_progress"`
+    StartTime  time.Time   `json:"start_time"`
+    EndTime    time.Time   `json:"end_time"`
+    // For additional info
+    Duration   int         `json:"duration"`
```
nlowe: What's the units on this? Seconds? Milliseconds? Do we even need to pre-calculate it if we give a start and end-time?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    Logs       []*TaskLog  `json:"logs"`
+}
+
+// The task log
+type TaskLog struct {
+    // The task ID
+    ID           string    `json:"id"`
```
nlowe: What's the format of this ID?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    ResourceName string    `json:"resource_name"`
```
nlowe: Is this the name of the source or destination resource? Do we need to track both?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    StartTime    time.Time `json:"start_time"`
+    EndTime      time.Time `json:"end_time"`
+    Status       string    `json:"status"`
```
nlowe: Should this be an enum instead?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+All the registered adapter implementation is maintained in the `Registry Adapter Registry` and utility methods provided to get the corresponding adapter by the adapter identity.
+
+```go
+// The type of adapter
+type AdapterType string
+
+// AdapterInfo provides base info and capability declarations of the adapter.
+type AdapterInfo struct {
+    // UUID of the adapter
+    Identity string                       `json:"identity"`
```
nlowe: For consistency, this should be the `ID` of the `AdapterInfo`. Also, why is this a `UUID`?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    Info()*AdapterInfo
+    // Get all the available namespaces under the specified registry with the
+    // provided credential/token.
+    GetNamespaces()([]Namespace, error)
+    // Create a new namespace
+    CreateNamespace(ns Namespace) error
+    // Fetch the content resource under the namespace by filters
+    // SUGGESTION: Adapter provider can do their own filter based on the filter pattern
+    // or call the default `DoFilter` function of the filter to complete resource filtering.
+    FetchResources(ns Namespace, filters []*Filter)([]*Resource, error)
```
nlowe: Why do we need to fetch resources for adapters?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+
+// Adapter interface defines the capabilities of registry
+type Adapter interface {
+    // Info return the information of this adapter.
+    Info()*AdapterInfo
+    // Get all the available namespaces under the specified registry with the
+    // provided credential/token.
+    GetNamespaces()([]Namespace, error)
+    // Create a new namespace
+    CreateNamespace(ns Namespace) error
```
nlowe: Does this need to be idempotent? If so, that should be documented.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+    Name string                           `json:"name"`
+    // One sentence to describe the adapter
+    Description string                    `json:"description"`
+    // The supported resource types
+    SupportResourceTypes   []ResourceType `json:"support_resource_types"`
+    SupportResourceFilters []FilterType   `json:"support_resource_filters"`
+}
+
+// Adapter interface defines the capabilities of registry
+type Adapter interface {
```
nlowe: Can you give an example of what an implementation for this would look like, say for AWS or Azure's container registries?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+This transfer job is simple, get the corresponding transfer handler to complete the data transfer process based on the input resource.
+
+## Non-Goals
+
+* This proposal only covers the replication base framework design, not include the designs of each concrete adapters for kinds of registries.
+* This proposal only provides the main content of the replication NG, not contain all the details of the service/components. The missing detail can be provided in the concrete implementation code.
+
+## Rationale
+
+[None]
```
nlowe: Nit: Sections of the template that do not have any comments can be omitted.

proposals/Content-Replication-ng.md
steven-zou: Get all the content (image repository/chart) under the namespace may have differences among different registry providers (not covered in the registry V2 API), so, the adapter needs to take care of that difference.

proposals/Content-Replication-ng.md
steven-zou: +1

proposals/Content-Replication-ng.md
steven-zou: Removed

proposals/Content-Replication-ng.md
steven-zou: Removed

proposals/Content-Replication-ng.md
steven-zou: Providing the example you figured out may take many efforts, I don't think it's necessary if we well defined the interface.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,653 @@
...
+The replication NG should cover the following use cases:
+
+* User can manage and monitor both Harbor and non-Harbor registry endpoints
+* User can create replication policies to trigger replication
+* User can replicate both image and helm charts content
+* User can replicate filtered content (may be from multiple sources) from upstream registry (Harbor or non-Harbor) to current Harbor registry `pull-based`
+* User can replicate filtered content in current Harbor registry to the downstream registry (Harbor or non-Harbor) `push-based`
+* User can check the replication summary report
+* User can track and control (stop/cancel/pause?/resume?) the replication process
+* User can use replication cross-geo to sync content with reasonable performance
```
mmpei: so does this prosoal include a performance plan?

proposals/Content-Replication-ng.md
nlowe: But it's a breaking change. Are we going to keep the existing API surface available until a major version bump?

proposals/Content-Replication-ng.md
nlowe: So it's more of an `AdapterType`?

proposals/Content-Replication-ng.md
nlowe: Ok, but why do we need to know about that for replication in the first place? If the content already exists wouldn't it be a NOP anyways?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,678 @@
...
+  DELETE "/api/targets/:id([0-9]+)"           // Remove the specified registry endpoint
+  GET    "/api/targets/:id([0-9]+)/policies/" // List registries by registry
+  ```
+
+  * **NG API route definitions**:
+
+  ```go
+  /**
+  * 1. As the managed registries may be sources or targets in replication NG, the `targets` term in the API routes of registry management is not suitable anymore. Use `registries` to instead.
+  * 2. As the health checking is handled by the backend monitor, ping can be moved out.
```
mmpei: if ping moved how to check whether a registry is available before adding?

proposals/Content-Replication-ng.md
mmpei: i think number is enough for the id of registry couldn't exceed max value of long in normal.

proposals/Content-Replication-ng.md
steven-zou: @nlowe 
I’m thinking the same question. Seems keeping the existing ones for a while may be better for compatibilities. I'm ok with this.

proposals/Content-Replication-ng.md
steven-zou: @nlowe 

Not get your meanings well. The resources are located in the source registry which is managed under the specified namespace. We need to fetch them and upload them to the dest registry. I really don't understand `If the content already exists`.

proposals/Content-Replication-ng.md
steven-zou: No, we don't. We just need to consider the performance when doing implementations. (writing high quality code ^_^)

proposals/Content-Replication-ng.md
steven-zou: The first check should happen in the creation logic cycle with the internal call, no api need to be exposed.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,678 @@
...
+  ```go
+  /**
+  * 1. Add `replication` as a API group context to indicate those APIs handle replication related things.
+  * 2. Each replication running based on the policy will be recorded as an `Execution`.
+  * 3. Move the DELETE replication execution out.
+  */
+  POST "/api/replication/executions"                 // Start/trigger a replication execution
+  GET  "/api/replication/executions"                 // List all the summaries of replication executions
+  GET  "/api/replication/executions/:id([0-9]+)"     // Get the specified execution detail
+  GET  "/api/replication/executions/:id([0-9]+)/log" // Get the log data of the specified replication execution
```
ywk253100: Need the job ID when trying to get the log?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,678 @@
...
+    // The key of the access account, for OAuth token, it can be empty
+    AccessKey string
+    // The secret or password for the key
+    AccessSecret string
+}
+
+// Namespace represents the full path of resource isolation unit;
+// if the namespace has hierarchical structure, e.g organization->team,
+// it should be converted to organization.team
+type Namespace string
```
ywk253100: The `namespace` also has its own metadata, such as `public/private` in Harbor

lweitzman:
Screenshots of mockups attached and mockups of those images can be seen here: https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/340549431_v1-8_-_Registries 
image002
image003
image004
image005

proposals/Content-Replication-ng.md
steven-zou: That's right. So I define a new type to cover it. I just ignore what methods it should have. Let the component owner handle it.

proposals/Content-Replication-ng.md
steven-zou: Add id for the logs

steven-zou:
> Screenshots of mockups attached and mockups of those images can be seen here: https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/340549431_v1-8_-_Registries
> image002
> image003
> image004
> image005

@lweitzman 

Thanks louis for the UI mockup. About the replication policy list view, I did not provide mockup in this proposal, but it has some minor changes. More info columns should be included in the list view.

- src registry
- dest registry
- src namespaces
- dest namespaces

steven-zou:
@nlowe 

Any other comments to the main process? if no, please approve the PR. Then we can start the work.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+  * Do other necessary actions if necessary.
+* **Data Transferring**: Do the real data transfer action here. The data upstream and downstream are connected through an IO buffer to achieve the purpose of improving performance by reading and writing at the same time.
+* **Post Action**: Do some clean work if needed or send notifications to the interested parts after the replication completed.
+
+#### Resource
+
+To significantly simplify the process, a top level abstracted resource should be defined to make the resource object unified and consistent.
+
+The resource mentioned in this proposal refer to container image and Helm Chart. Those two kinds of resource can be delegated by the following resource abstraction:
+
```
reasonerjt: Header/Metadata/Additional Info/Extended

Seems these all fall into the category of `Metadata`

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+  DELETE "/api/targets/:id([0-9]+)"           // Remove the specified registry endpoint
+  GET    "/api/targets/:id([0-9]+)/policies/" // List registries by registry
+  ```
+
+  * **NG API route definitions**:
+
+  ```go
+  /**
+  * 1. As the managed registries may be sources or targets in replication NG, the `targets` term in the API routes of registry management is not suitable anymore. Use `registries` to instead.
+  * 2. As the health checking is handled by the backend monitor, ping can be moved out.
```
reasonerjt: `ping` is not only for health checking, it's also for credentials validation.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+  /**
+  * 1. Add `replication` as a API group context to indicate those APIs handle replication related things.
+  * 2. Each replication running based on the policy will be recorded as an `Execution`.
+  * 3. Move the DELETE replication execution out.
+  */
+  POST "/api/replication/executions"                  // Start/trigger a replication execution
+  GET  "/api/replication/executions"                  // List all the summaries of replication executions
+  GET  "/api/replication/executions/:id([0-9]+)"      // Get the specified execution detail
+  GET  "/api/replication/executions/:id([0-9]+)/logs" // Get all the task logs of the specified execution
+  GET  "/api/replication/executions/:id([0-9]+)/logs/:id([0-9]+)"  // Get the log text data of the specified replication execution log
```
reasonerjt: Does it imply an execution will be broken down into multiple tasks?

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+  * From which namespaces (src namespaces) to pull content
+  * What types of content will be pulled
+  * What filters can be applied to narrow down the replication (name, version and labels?)
+  * Push content to which registry (dest registry)
+  * Push content to which namespaces
+  * How to trigger this policy
+    * Schedule a time
+    * Event based (push/delete)
+    * Manual (not automatically executed)
+* Policy modes
```
reasonerjt: Seems this `policy mode` overlaps with `How to trigger`?

nlowe:
@steven-zou There are a few comments from my initial review (hidden behind the `19 more items`) that'd I'd love your input on. I'll try to take another pass today and see if I missed anything.

proposals/Content-Replication-ng.md
steven-zou: To explicitly declare what function for each property section and different property section has different stabilities

proposals/Content-Replication-ng.md
steven-zou: The validation is handled backend and reflected by the status field. The validation should be down when creating/updating and continuously updated by the loop monitor.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+
+The flow is logically splitted into three stages: prepare, data transferring and post-actions:
+
+* **Prepare**: All the preparations work is happened here.
+  * The connectivity of upstream/downstream registries will be checked to achieve a early failure in case there is something wrong;
+  * The replicating resources are retrieved with filter policies if set from the upstream registry;
+  * Prepare the downstream registry replicating environment, most cases are creating corresponding namepsaces;
+  * Do other necessary actions if necessary.
+* **Data Transferring**: Do the real data transfer action here. The data upstream and downstream are connected through an IO buffer to achieve the purpose of improving performance by reading and writing at the same time.
+* **Post Action**: Do some clean work if needed or send notifications to the interested parts after the replication completed.
```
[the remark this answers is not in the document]
steven-zou: Maybe, not confirmed. Here just list out some potential cases can be done in such a stage. We may not provide any functions here so far as discussed in the WG.

proposals/Content-Replication-ng.md
steven-zou: Right. Executed by multiple tasks but track results by a unified object.

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,681 @@
...
+
+#### Policy Management
+
+The policy is actually a data model to define the replication template, the policy manager will cover the **CURD** operations of that policy data model.
+
+* The policy should define:
+  * From which registry (src registry) to pull content
+  * From which namespaces (src namespaces) to pull content
+  * What types of content will be pulled
+  * What filters can be applied to narrow down the replication (name, version and labels?)
```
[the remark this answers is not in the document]
steven-zou: Good point. The filters are open for extending.

proposals/Content-Replication-ng.md
steven-zou: Sure

proposals/Content-Replication-ng.md
steven-zou: ok

proposals/Content-Replication-ng.md
steven-zou: To indicate if replication the deletion ones (actually if replay the deleting action in the target side)

proposals/Content-Replication-ng.md
steven-zou:
```go
Filter{
  Type: "Name",
  Value: "*_test*"
}
```

The registry adapter provider can decide how to use this information (how to match the pattern), we will not limit the operation as we did before. (Currently, the filter is an interface with `DoFilter` function, we implement different filters and call those filters by the chain on the resources we have fetched from the src.  But currently, we're considering some registry provider may support appending query parameters after their resource fetching API)

proposals/Content-Replication-ng.md
steven-zou: Give admins more info about the endpoints. Not a mandatory thing, just a fancy thing.

proposals/Content-Replication-ng.md
steven-zou: I originally planned to use string uuid to instead the int id (from db table) in the new code. To avoid confusions, let's keep as it is.

proposals/Content-Replication-ng.md
steven-zou: Actually, the `Insecure bool          `json:"insecure"`` means skipverifycert, not mean support HTTP registry.

proposals/Content-Replication-ng.md
steven-zou: Right, the controller just provides related interface methods to the restful API layer and `flow controller` primarily driver the flow defined before (most work is covering the preparation work)

proposals/Content-Replication-ng.md
steven-zou: New one, it has different function/responsibilities with the existing scheduler. It'a replication scheduler, mainly do communications with job service (Request jobs).

proposals/Content-Replication-ng.md
steven-zou: Added

proposals/Content-Replication-ng.md
steven-zou: Just publish replication job execution results to the replication manager to let it update the execution report.

proposals/Content-Replication-ng.md
steven-zou: Let's keep the same with other models to avoid confusions.

proposals/Content-Replication-ng.md
steven-zou: Let's keep the same with other models to avoid confusions.

proposals/Content-Replication-ng.md
steven-zou: Good point. Let's have a try.

proposals/Content-Replication-ng.md
steven-zou: Status? Both are ok for me. Let the feature developer/owner to determine the details. The proposal may not touch too detailed code level thing. All the code here is just to indicate the main idea of the proposal, they're not the finalized code.

proposals/Content-Replication-ng.md
steven-zou: No concrete cases, so far. Just use it as a placeholder to provide more flexibilities to the struct. The main possible use case brings some info/data from some very specified registries.

proposals/Content-Replication-ng.md
steven-zou: Not give all the details of the `Namespace`, let the feature developer/owner to design and determine what functions should namespace has, is that ok?

steven-zou:
> @steven-zou There are a few comments from my initial review (hidden behind the `19 more items`) that'd I'd love your input on. I'll try to take another pass today and see if I missed anything.

@nlowe I saw that and comments to your comments.

steven-zou:
Aggregate and link the previous PRs of replication related to this one:

https://github.com/goharbor/community/pull/21
https://github.com/goharbor/community/pull/26

proposals/Content-Replication-ng.md
```diff
@@ -0,0 +1,687 @@
...
+
+The `Replication Flow Controller` accepts the replication call from the `Controller` and start to initialize the flow process. Get the policy from the `Policy Manager` and parse it and inject into the flow. Get the registry data from the `Registry Endpoint Manager` and wrap data into the flow. Base on the above data, retrieve and constructed related registry adapters from the `Adapter Manager`. Retrieve resources (image/chart/all) from the specified source registry and apply filter polices to get the replicating resource candidates.
+
+The `Scheduler` schedule jobs to do the real data transferring of the resource candidates generated by the `Replication Flow Controller` by calling job service APIs. The `Scheduler` should follow the below interface:
+
+```go
+// Schedule tasks to transfer resource data
+type Scheduler interface {
+    // Schedule tasks
+    Schedule(srcResources []*Resource, destResources []*Resource) error
```
nlowe: Are these always the same length such that `srcResources[i]` is replicated to `destResources[i]`?

steven-zou: Not sure, depends on what we'll handle the case that the same image repos from different src NS are replicated to the same target NS.

PR #50, opened 2019-02-21
Change http to https for security links

huynq0911:
For security, we should change http into https links.


huynq0911:
> @huynq0911
> 
> Seems the `https` for the wiki page does not work well as the `HTTP` link.

hmm, I see. The layout is broken. I will close this PR. Thanks!

PR #52, opened 2019-02-21
Update MEETING_SCHEDULE.md

michmike:
because of daylight savings that ends on March 10 2019, the timezone conversions from China to USA are not accurate


michmike:
my PR was messed up. will recreate it

PR #55, opened 2019-02-22
Update MEETING_SCHEDULE.md

michmike:
updating the meeting schedule to include the correct timezones for US/China (because of daylight savings the translation between timezones is not always accurate).
also included a google calendar that can be imported

Signed-off-by: Michael Michael <michmike@users.noreply.github.com>


michmike:
my inability to sign my PRs is troubling :( i have tried every way imaginable but my windows laptop is not letting me. i will try once more after this.

PR #62, opened 2019-03-08
Add proposal for CLI password

reasonerjt:


reasonerjt:
Merging it so we can work together to refine the proposal

PR #63, opened 2019-03-08
[Draft]: CLI password proposal 

reasonerjt:


lweitzman:
Mockups here: 
https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/325339955_Quotas_-_Users
An administrator can see which users have a generated password and can revoke them. In addition, the menu in the upper right (in addition to reseting a password) can now generate a cli password for any user.

nlowe:
While this would be useful in my environment, how is this different from #35, which has already been merged? As far as I can tell, the only difference here is we're giving a way for an individual to log in without using their password for use in a CI-like environment. Doesn't this duplicate most of the functionality of #35?

ericvmw:
I have a concern on the lack of expiration support in the CLI password proposal. It is not uncommon to require user password expiration support for security compliance in on-prem deployments. Though this is a generated password, it is used by end users. I would strongly request the expiration support on the CLI password. When it expires, the end user can renew it to get a new one.

reasonerjt:
@nlowe , the difference between `cli password` and `robot account`, is that the permission for `cli password` is same as the user and it's the regular user who generates the `cli password`, while `robot account` is managed by the admin, i.e. they have different scope.

Per my understanding both `robot account` and `cli password` solve your problem.  Another important usage for `cli password` is for the user onboarded via OIDC provider, we need a solution for the oidc user to use `docker` cli while the `docker` cli does not support the SSO redirection flow.

For more details about OIDC integration, a PR will be opened for the draft of proposal with in a couple of days.

steven-zou:
Two concerns:
* The authN of docker CLI is handled by Harbor itself with the saved secret in its database no matter if I configured a backend identity provider. That means there will be two auth systems working at the same time.
* The function of the CLI password is very similar to the robot account. Adding a new auth approach will cause more messes and confusions.

reasonerjt:
@steven-zou 
1. What's the biggest issue for the two systems?  The other system is created intentionally so Harbor can verify the user given the limitation of the CLI.
2. There is some overlap but I don't think it will really confuse user, `robot` account is for project admin, a regular user doesn't have permission to create it.

lweitzman:
updated mockups to move the user management of the CLI password into the User Profile dialog. mockup here: https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/325042732_Folders_-_Collection

reasonerjt:
I'm closing this pr as we don't want to introduce this feature.

PR #64, opened 2019-03-12
Proposal for OIDC support

reasonerjt:


proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+he should be logged in with the onboarded username. 
+
+**NOTE:**
+There are other attributes in user's profile, such as email, it is not modifiable during onboard process, the code will try 
+to use the claim in the token to fill in such values and user can modify it after onboard.  The username is not modifiable once 
+is onboarded.
+
+* #### User onboarded via the OIDC authentication flow accessing the API.
+
+After a user is noboarded via OIDC authentication flow and assigned proper permissions, he should be able to trigger API
```
ktbartholomew: s/noboarded/onboarded

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+cannot verify the password in the request.
+
+* #### User onboarded via the OIDC authentication flow using docker/helm CLI 
+After a user is onboarded and assigned proper permissions, he should be able to  use docker/helm CLI to pull/push artifacts 
+from/to Harbor.
+There's a gap in this scenario given the fact that the CLI tools does not support redirection, that user cannot authenticate 
+against the external OIDC provider.  To avoid adding extra dependencies to user, we don't want to create another CLI tool.
+There are some possible solutions as follow:
+
+1. Provide a _CLI password_ mechanism.  After a user is onboarded via OIDC, he has to create a CLI password and use this
```
ktbartholomew: I think this is the most viable option described here. The biggest caveat is that a CLI password could continue to be used well after a user's OIDC token has expired, and perhaps after that user has been deactivated by the OIDC provider. With the `offline_access` scope, and therefore a refresh token, we _might_ be able to validate that the user is still active with the OIDC provider. I think there are some cases in which Dex doesn't support refresh tokens, so this may not always work. Additionally, this would mean we would need to store a user's refresh token in the database.

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+* Integration with none-OIDC identity managers such as Keystone
+* Support multiple identity managers, such as OIDC/LDAP, OIDC/DB at the same time.
+
+## Compatibility
+
+For verification, we'll need to ensure it works with the following OIDC providers:
+* Google, [dex](https://github.com/dexidp/dex) -- P0
+* [Keycloak](https://github.com/keycloak/keycloak) -- P1
+
+## Implementation
```
ktbartholomew: We'll also need some way to store and validate the `state` parameter for the OAuth flow to prevent CSRF attacks. Could this just be stored in the user's session? Along with state, we would also need to store the "redirect to" path, so the user is sent to the correct deep link after the OAuth flow is complete.

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+1. Provide a _CLI password_ mechanism.  After a user is onboarded via OIDC, he has to create a CLI password and use this 
+password and his username to authenticate while using the docker/helm CLI.  Some more details in the proposal:
+https://github.com/goharbor/community/pull/63
+2. Use OIDC token as password, or expose some random string as "secret" to map to the OIDC token, so the OIDC token is verified
+in the authentication process when using CLI.  The problem with this approach is that OIDC token normally expires in hours, after 
+that user will have to login to Harbor again to get a new token.  
+
+## Non-Goals
+
+* Integration with none-OIDC identity managers such as Keystone
```
ktbartholomew: s/none/non

proposals/OIDC_Authentication_Suppot.md
ktbartholomew: I will do some additional research to see if this refresh token technique is viable.

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+
+After setting the values the Administrator should be able to test the connection to the OIDC provider and verify the 
+client ID and Secret.
+
+**There will be some limitations:**
+* Minimized attributes: Some OIDC providers may require additional attributes in configuration, for them the flow may not 
+work.
+* Only one OIDC endpoint will be supported:  There is some limitation in Harbor's configuration management that there is
+no good way to support list as a configuration group.
+* Hard coded scope: We will not allow admin to customize the scope, it will be hard-coded to `openid`
```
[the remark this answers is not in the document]
ktbartholomew: I confirmed that when `profile` scope is not requested, Dex does not return any username info. This would force the user to provide their own username during onboarding, with a 0% chance that their preferred username is already pre-filled. I think the user experience would benefit a lot from having that value pre-filled from whatever identity store is backing the OIDC provider, so we should try to include the `profile` scope if the provider supports it.

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+I propose we make enhancement so that a unified workflow for user to configure and integrate external OIDC providers with 
+Harbor.  Admin can configure an OIDC provider, after that, user can use this OIDC provider as authentication
+approach to kick off an SSO flow to log into Harbor's UI Portal.  Besides that, user should be able to use a valid OIDC token
+to access Harbor's API, the user should also be able to use docker and helm CLI.
+
+More details for each major use case as follow:
+
+* #### Administrator Configuring OIDC Endpoint
+
+The Administrator of Harbor should be able to configure the OIDC Provider via Portal or configuration API.
```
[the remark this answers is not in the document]
reasonerjt: Yes, admin `user_id==1` can always login regardless the auth setting.

proposals/OIDC_Authentication_Suppot.md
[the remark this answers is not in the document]
reasonerjt: I think it justifies the "scope" as an attribute of the OIDC auth setting.  By default it should be `openid, profile, email`.

proposals/OIDC_Authentication_Suppot.md
reasonerjt: IMO, if we need to do some validation on the OIDC provider, we have to tell the user that he has to do something on his browser to input his credentials on the OIDC provider's login page and get the latest OIDC token or some sort of secret, use it as password in CLI.  The password/secret will be short-lived.

proposals/OIDC_Authentication_Suppot.md
```diff
@@ -0,0 +1,110 @@
...
+The OIDC authentication will be exclusive to other authentication methods, i.e. when admin set to OIDC authentication, 
+user registration will be disabled.
+
+The attributes of OIDC provider:
+
+* **Name**: This is the name of the OIDC provider.  It has to be https protocol and is a required attribute.
```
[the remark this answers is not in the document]
reasonerjt: I was thinking the `Name` attribute should be helpful for UI to render the login link like `Login via OIDC provider: ${Name}`.
And yes the protocol should be moved to "Endpoint URL".

proposals/OIDC_Authentication_Suppot.md
reasonerjt: by `redirect to` do you mean the redirect URL that was set when registering the client on ODIC provider?  I don't think it should be configurable as it has to be the same as was registered.

Yes storing the `state` in the session should work, the state should be generated when redirecting to the OIDC login page and verified by the code handles the `redirect URL`.

proposals/OIDC_Authentication_Suppot.md
reasonerjt: Thanks!

proposals/OIDC_Authentication_Suppot.md
reasonerjt: Thanks!

proposals/OIDC_Authentication_Suppot.md
ktbartholomew: As I was digging around the Dex documentation, I found that it does not always support refresh tokens. For instance, when SAML is used as the connector refresh tokens are never sent to the client. Their reasoning is that (similar to our current problem) Dex would have no way of communicating with the backend to determine if a user still exists. So there may be scenarios in which we have no choice but to manage a secret CLI password locally within Harbor. Pursuing both options is probably overly burdensome to develop, and potentially confusing for users. I suppose we should just stick with the CLI password.

Should we expose a configuration option for the expiry of the CLI password? Some users may be fine with passwords that rarely expire, whereas others may insist on better security at the expense of user experience.

proposals/OIDC_Authentication_Suppot.md
ktbartholomew: > by redirect to do you mean the redirect URL that was set when registering the client on ODIC provider?

No, I mean that if a user tried to access `/harbor/projects/library` before starting the authorization flow, they should be sent back to that page when the authorization flow is complete.

proposals/OIDC_Authentication_Suppot.md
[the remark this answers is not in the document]
reasonerjt: @ktbartholomew I think this comment can be resolved?

proposals/OIDC_Authentication_Suppot.md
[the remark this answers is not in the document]
reasonerjt: @ktbartholomew I have updated the PR, I believe we can resolve this comment?

proposals/OIDC_Authentication_Suppot.md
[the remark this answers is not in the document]
reasonerjt: I've updated the PR, if there no disagreement I'll resolve it.

@ktbartholomew Any concerns?

proposals/OIDC_Authentication_Suppot.md
reasonerjt: I think the key point in the debate, is that to verify whatever secret that associates to a OIDC onboarded user, it should verify against the OIDC endpoint, not only on Harbor.  Otherwise, if the user is invalidated on the OIDC provider, the secret may still work for a long time.

How about we solve it like this:  
When a user is onboarded, a `secret` is created and associated to both `id_token` and `refresh_token`.  This secret will be used as the CLI password for `docker login`.  And when the auth backend is set to OIDC, Harbor will first try to verify the `id_token` associated with the `secret`, if the `id_token` is expired it tries the `refresh_token` and updates the `id_token` and `refresh_token`.
When user is logged out the `secret` will not be removed, such that the flow above still works.

For the case when Dex is connected with SAML 2.0, we will document that refresh token will not work, and user will have to re-login via UI, such that the `secret` can be associated with a new `id_token` such that the CLI flow still works.

I briefly checked the doc, only potential challenge is that different OIDC providers may have different ways for user to get `refresh_token`.  For example, in Google user needs put `access_type=true` in the request, and in Dex user needs to put the `offline_access` in the scope.  We'll double check if there's a way to generate authn request that works for both backends, if it works, this is a preferred way than `CLI password`

Thoughts?

proposals/OIDC_Authentication_Suppot.md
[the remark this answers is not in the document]
ktbartholomew: agreed. I don't think I have permission to resolve it, though.

proposals/OIDC_Authentication_Suppot.md
reasonerjt: I see, I think we can associate the state to the `redirect to` URL if user was kicked back to login window when he's accessing a specific URL of Harbor.

steven-zou:
Per lazy consensus rule, merge this PR to main.

PR #81, opened 2019-06-10
Add proposal of project quota

wy65701436:


proposals/new/project-quota.md
```diff
@@ -0,0 +1,200 @@
...
+CREATE TABLE Manifest (
+ id SERIAL PRIMARY KEY NOT NULL,
+ digest varchar(255) NOT NULL,
+ digest_blob varchar(255) NOT NULL
+);
+```
+
+Table -- Project Quota Usage
+
+```
```
reasonerjt: Could you add more comments about each column of the table?

proposals/new/project-quota.md
```diff
@@ -1,122 +1,168 @@
...
 ### Story 4
-As a system administrator, I can enlarge the project quota per request, either on storage usage or artifact count.
-
+As a system administrator, I can set the project quota, either on storage usage or artifact count.
+​
 ### Story 5
 As a system administrator, I can view all project quota metrics via a dashboard.
 
+### Story 5
+As a system administrator, If I update the default quota, the existing project won't be impacted.
```
renmaosheng: adding "only the new created project will inherit the new default value"

proposals/new/project-quota.md
```diff
@@ -1,122 +1,168 @@
...
 ## Proposal
-
-I propose the following solutions:
-
+​
+We propose the following solutions:
+​
 1. Add quota management page in administration so that system administrators can set storage and number quota for the project.
 2. Add quota information page in the project to display the resources used by the project for project administrators.
 3. Modify or hook the docker distribution proxy handler in core component, denied push when new pushed image over the storage and number quota.
```
renmaosheng: Could also add some more details like push (e.g. push image, replication, etc?)

proposals/new/project-quota.md
```diff
@@ -1,122 +1,168 @@
...
-4. After image uploaded or deleted, update the storage usage of the project, which will be used in step 4.
+4. After image uploaded or deleted, update the storage usage of the project.
```
renmaosheng: it is better to consistently using push/delete, both storage usage and counts will be updated.

proposals/new/project-quota.md
```diff
@@ -1,122 +1,168 @@
...
-### Project Quota Setting
-
-## API
-
- 1. List quotas
-
+ 1. List quotas of the projects
     ```
-    GET /api/quotas
+    GET /api/project-quotas/
```
renmaosheng: I assume this will be updated based on 6/18's discussion.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+
+```
+CREATE TABLE Manifest (
+ id SERIAL PRIMARY KEY NOT NULL,
+ digest_af varchar(255) NOT NULL,
+ digest_blob varchar(255) NOT NULL
+);
+```
+​
+## Data Flow in Docker registry
```
reasonerjt: The way you lock "project usage" will severely impact parallel push

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+ digest varchar(255) NOT NULL,
+ type varchar(255) NOT NULL,
+ CONSTRAINT unique_artifact UNIQUE (repo, tag)
+);
+```
+
+Table -- ArtifactAndBlob
+
+```
+CREATE TABLE Manifest (
```
reasonerjt: it should be relationship of IDs in this table.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+  //            WHERE content_type = ? AND object_id = ? AND resource = ? FOR UPDATE
+  // 2. UPDATE quota SET used = used WHERE id = ?
+}
+func NewQuotaManager(reference string, referenceId string) *QuotaManager {
+  // ...
+}
+
+```
+
+​​
```
reasonerjt: Will there be a general framework for quota checking and how is it invoked?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+CREATE TABLE Manifest (
+ id SERIAL PRIMARY KEY NOT NULL,
+ digest_af varchar(255) NOT NULL,
+ digest_blob varchar(255) NOT NULL
+);
+```
+​
+## Data Flow in Docker registry
+​
+The date flow to push a image into Harbor:
```
reasonerjt: You only need to lock once for both checks?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+​
+  ```
+  POST /api/internal/dumpregistry
+  ```
+​
+## Consideration of performance
+1, Docker push.
+> Redirect the HEAD request to DB before pushing a blob.
+​
+2, Registry Client in Core.
```
reasonerjt: Please clarify which API call from registry Client can be replaced by querying DB.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+But, if we count size for blob, all of them may fail.        
+​
+## API
+ 1. Registry Dump (Sysadmin only)
+​
+  ```
+  POST /api/internal/dumpregistry
+  ```
+​
+## Consideration of performance
```
reasonerjt: Need to evaluate the performance gain before taking action.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+> Replace the API by calling DB, like get manifest, layer digest, image size.
+​
+## Non-Goals
+​
+# Docker images
+Don't split the shared image layer size into pieces, each shared layer will count its size into the total size of a project.
+It causes the total usage of a Harbor instance is not reflect true value.
+​
+## Compatibility
+It has to consider how to handle the migration from older version, like v1.7.0.
```
reasonerjt: We MUST have a very clear migration plan.
Like, how do we expect user to trigger the migration script in the upgrade flow.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+```go
+
+type QuotaManager struct {
+  reference string
+  referenceId string
+}
+
+func (m *QuotaManager) RequestResource(resource string, value int64) error {
+  // 1. SELECT * FROM quota 
+  //            WHERE content_type = ? AND object_id = ? AND resource = ? FOR UPDATE
```
ywk253100: I think we should use the optimistic concurrent control rather than the pessimistic one?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,252 @@
...
+### Failure cases
+1, Storage overuse. If we let the all of blobs pass the flow, the real storage usage could larger than quota usage.
+
+  
+The execution flow to put a blob:
+
+  ![Blob Push flow](../images/project-quota/flow_blob.png)
+  
+### Failure cases 
+1, Quota overuse. If 1 repo number quota left, Harbor could let only one image pass in multiple push scenario, but other blobs already count into the usage.
```
ywk253100: I don't think this is a failure case, it's the right behavior of quota.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,252 @@
...
+
+2, If 100 MB storage quota left, there are 3 push in parallel, 70M, 90M, 20M. In fact, Harbor should let at least one of them pass.
```
ywk253100: Same comment with case 1.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
+hweiwei
+
+
+
+
```
alexvmw: add

proposals/new/project-quota.md
```diff
@@ -0,0 +1,200 @@
+# Project Quota
+
+Author: He Weiwei/Yan Wang
+
```
alexvmw: add

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+​
+1. Add quota management page in administration so that system administrators can set storage and number quota for the project.
+2. Add quota information page in the project to display the resources used by the project for project administrators.
+3. Modify or hook the docker distribution proxy handler in core component, denied push when new pushed image over the storage and number quota.
+4. After image uploaded or deleted, update the storage usage of the project.
+5. Modify or hook helm chart upload API, denied upload when new uploaded chart over the storage and number quota.
+6. After helm chart uploaded or deleted, update the storage usage of the project.
+7. Recalculate the storage usage by project whe upgrade from old version.
+​
+​
```
alexvmw: can you check quota behavior for these basic scenarios? @reasonerjt @jwangyangls 

Per project quotas configured – 
1) 10 images max 
2) 100GB disk space max

Assume right now, said project contains 8 images and 80GB disk space used

Assume two images (imageA, imageB) are pushed concurrently.  Assume first that there is no shared layers among them. Consider the following scenarios

ImageA - 5gb & imageB -10gb 
Result: both will be pushed successfully (we don’t consider failure of pushing individual layers, assume this is always successful).  

ImageA – 10gb & imageB – 15gb
Result: 
1.	either one of the two images will succeed (imageA or imageB pushed fully along with manifest before quota is hit) or 
2.	neither will succeed (both images are only partially pushed before quota hit)

now assume imageA and imageB have shared layers (where each layer is uniformly 1 gb)

ImageA - 5gb & imageB -10gb 
Result: both will be pushed successfully no many how many layers shared

ImageA – 10gb & imageB – 15gb
Result:	
1.	 if >= 5 layers shared then, both will be pushed successfully
2.	 if <5 layers shared, either
a.	Either one of the two images will succeed (imageA or imageB pushed fully before quota hit) or
b.	Neither will succeed (both images pushed partially when quota hit)

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Story 5
+As a system administrator, If I update the default quota, the existing project won't be impacted.
+​
+​
+## Proposal
+​
+We propose the following solutions:
+​
+1. Add quota management page in administration so that system administrators can set storage and number quota for the project.
```
alexvmw: 1. limit by size(gb), 2.limit by number, are these two and / or? admin can set only one of the two or set both correct? @wy65701436

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+As a system administrator, I can set the default limit on project artifact count and project storage. But, I cannot control how many projects can Harbor have.
+​
+### Story 2
+As a project administrator, I can only view basic quota information in project quota page, which includes quota usage, limits and errors. 
+​
+### Story 3 (OUT)
+As a project administrator, I can exceed my project limit by sending the request to system admin and wait for response.
+​
+### Story 4
+As a system administrator, I can set the project quota, either on storage usage or artifact count.
```
alexvmw: or both?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+​
+### Story 5
+As a system administrator, I can view all project quota metrics via a dashboard.
```
alexvmw: @lweitzman please think about this

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Story 5
+As a system administrator, If I update the default quota, the existing project won't be impacted.
+​
+​
```
alexvmw: want to add a 
Story 6: given how easily it is to hit quota when you concurrently push images to a project, we want some kind of warning (maybe leveraging Webhook) when usage hits a certain %, maybe user-configurable since he is more knowledgeable about his pushing behavior.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,255 @@
...
+  ![Docker Push flow](../images/project-quota/data_flow_registry.png)
+  
+
+The execution flow to put a manifest:
+
+  ![Manifest Push flow](../images/project-quota/flow_manifest.png)
+  
+### Failure cases
+1, Storage overuse. If we let the all of blobs pass the flow, the real storage usage could larger than quota usage.
+
```
alexvmw: what happens then? if only a subset of the layers come through but not all and manifest doesn't get pushed through, are these residuals cleaned by calling GC?

proposals/new/project-quota.md
wy65701436: or and both. ok, will be updated.

proposals/new/project-quota.md
wy65701436: done.

proposals/new/project-quota.md
wy65701436: added

proposals/new/project-quota.md
wy65701436: always updated base on the latest implementation.

proposals/new/project-quota.md
wy65701436: defined an quota manager to manager quota reqire and release.

proposals/new/project-quota.md
wy65701436: for blob, it locks once. but for manifest, it locks twice.

proposals/new/project-quota.md
wy65701436: it's out of scope, but it could be considered in the future, like HEAD blob or manifest.

proposals/new/project-quota.md
wy65701436: will compare the report with the build that without quota middlewares.

proposals/new/project-quota.md
wy65701436: updated Compatibility part.

proposals/new/project-quota.md
wy65701436: in the Db, it's row lock, one kind of optimistic.

proposals/new/project-quota.md
wy65701436: no limitation in the front end. and system admin can set them both.

proposals/new/project-quota.md
wy65701436: story 2 is view, and story 5 is edit.

proposals/new/project-quota.md
wy65701436: yes. these are untagged blobs and could be released by GC and the usage could be updated in GC job.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+## User Stories
+
+### Story 1
+As a system administrator, I can set the default limit on project artifact count and project storage. But, I cannot control how many projects can Harbor have.
+
+### Story 2
+As a project administrator, I can only view basic quota information in project quota page, which includes quota usage, limits and errors. 
+
+### Story 3 (OUT)
```
nlowe: What does `OUT` mean in this context?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Story 7
+As a system administrator, I can choose the whether to use the default or customized quota on creating a project.
+
+### Story 8
+As a user, I cannot see the default quota on creating a project, but I can get it in the quota tab.
+
+## Premise
+1, The shared blob in project won't be be double-counted.
+2, The shared blob cross project will be be double-counted.
```
nlowe: Is this a technical limitation or is there a reason we're explicitly double-counting things?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+As a system administrator, If I update the default quota, the existing project won't be impacted.
+
+### Story 7
+As a system administrator, I can choose the whether to use the default or customized quota on creating a project.
+
+### Story 8
+As a user, I cannot see the default quota on creating a project, but I can get it in the quota tab.
+
+## Premise
+1, The shared blob in project won't be be double-counted.
```
nlowe: nit: This won't render as a numbered list on GitHub, it needs to be a `.` following the number, not a `,`

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+ digest_af varchar(255) NOT NULL,
+ digest_blob varchar(255) NOT NULL
+);
+```
+
+## Data Flow in Docker registry
+
+#The date flow to push a image into Harbor:
+
+  ![Docker Push flow](../images/project-quota/data_flow_registry.png)
```
nlowe: nit: relative paths for these are off by one level, you need an extra `../`

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+CREATE TABLE ArtifactAndBlob (
+ id SERIAL PRIMARY KEY NOT NULL,
+ digest_af varchar(255) NOT NULL,
+ digest_blob varchar(255) NOT NULL
+);
+```
+
+## Data Flow in Docker registry
+
+#The date flow to push a image into Harbor:
```
nlowe: nit: You need a space between the header level and title. Also, these probably shouldn't be H1's.

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+  ![Docker Push flow](../images/project-quota/data_flow_registry.png)
+
+
+#The execution flow of putting a manifest:
+
+  ![Manifest Push flow](../images/project-quota/flow_put_manifest.png)
+
+#The execution flow of putting a chart:
+
+  ![Manifest Push flow](../images/project-quota/flow_put_chart.png)
```
nlowe: nit: Alt-text should probably read:

```suggestion
![Chart Push Flow](../../images/project-quota/flow_put_chart.png)
```

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+    
+
+#The execution flow to putting a blob:
+
+  ![Blob Push flow](../images/project-quota/flow_put_blob.png)
+  
+The details to handle blob:
+1, for patch blob request, it records the size which is in the response header(Range) in redis on complete
```
nlowe: nit: This numbered list won't render correctly, Github expects a `.` after the number, not `,`

proposals/new/project-quota.md
nlowe: Is the intent to leave how this will be resolved to implementation, or should that go in the proposal as well?

proposals/new/project-quota.md
```diff
@@ -0,0 +1,367 @@
...
+		referenceID: referenceID,
+	}, nil
+}
+```
+
+There is a `Driver` interface for reference object which want to implement quota.
+
+```go
+// RefObject type for quota ref object
+type RefObject map[string]interface{}
```
nlowe: I'm not sure I'm seeing the technical need for the way this type is defined. This feels like a ***really*** weird way to program around some dynamic data / type restrictions. I Realize this already shipped in 1.9, did it get implemented differently?

renmaosheng:
It seems this proposal has not been merged yet, could you please merge it？thanks

hainingzhang:
LGTM

steven-zou:
Merge this PR according to the [lazy-consensus](https://github.com/goharbor/community/blob/master/GOVERNANCE.md#lazy-consensus) rule

PR #83, opened 2019-06-13
Adds README.md for new scanning workgroup

zhill:
Per discussion with @steven-zou , this is a PR purely for a README to create the scanning workgroup.


steven-zou:
As it's readme about the scanning workgroup, I'll directly merge it now.

PR #84, opened 2019-06-18
Add proposal for supporting vulnerability whitelist

reasonerjt:

### DIFF
--- a/proposals/CVE_Whitelist.md
+++ b/proposals/CVE_Whitelist.md
@@ -16,13 +16,18 @@ the vulnerabilities and explicitly ignore some of them during the policy checkin
 
 ## Proposal
 
-We propose to introduce the "whitelist" of vulnerabilities to project policy to solve this problem.  The system or project admin
-can define a list of CVE, when checking the scan result to compare the the threshold setting in "Deployment Security" policy, the 
-vulnerabilities in the white list will be ignored, such that user can pull an image with certain vulnerabilities higher than the 
-threshold without having to update or disable the policy.
+We propose to introduce the "whitelist" of vulnerabilities (a.k.a CVE whitelist) to project policy to solve this problem.  
+The system or project admin can define a list of CVE, when checking the scan result to compare the the threshold setting in 
+"Deployment Security" policy, the vulnerabilities in the white list will be ignored, such that user can pull an image with 
+certain vulnerabilities higher than the threshold without having to update or disable the policy.
+
+Each whitelist can optionally have a expiration time.  Once it's expired, the CVEs in that list will not be filtered when 
+calculating the overall vulnerability level.
 
 **NOTE:**
-For v1.9.0 there will be a limitation of the number of CVEs in the whitelist, which is 40.
+Each project can only have one CVE whitelist.
+The support for expiration may be a P1 for v1.9 release depending on the date and resource.
+There's not a limit of the number of items the a CVE whitelist but we need to think about it when implementing the UI.
 
 In detail, it will involve the following use cases:
 
@@ -42,11 +47,11 @@ admin can be reflected at the project level.
 * #### Pulling image from project with vulnerability whitelist
 
 Upon receiving a request to pull an image, instead of querying the "scan overview" in Harbor's DB, Harbor will call Clair's 
-API to list all the vulnerability of this image and apply to whitelist to filter out some of them, and consolidate the result
-to get the overall vulnerability level.
+API to list all the vulnerability of this image and apply to whitelist, if it's not expired, to filter out some of them, 
+and consolidate the result to get the overall vulnerability level.
 
 **NOTE:**
-Making the switch from querying local DB to calling Clair's API may have some impact on performance.  Storing the full vulnerability 
+Making the switch from querying local DB to calling Clair's API may have some impact on performance.  Storing the full CVE 
 list of every scanned image is not very effective.  We'll have to measure the performance during the refactor and lower the
 impact.
 
@@ -72,7 +77,7 @@ For managing and viewing the system level whitelist, these APIs should be added:
 All Harbor users should have permission to Call this API to view the list of system level vulnerability whitelist:
 
 ```HTTP
-GET /api/system/vuln_whitelist
+GET /api/system/CVEWhitelist
 Authorization: xxxxxxx
 ```
 
@@ -82,10 +87,16 @@ Authorization: xxxxxxx
 200 OK
 Content-Length: <length>
 Content-Type: application/json
-[
-    "CVE-2019-12310",
-    "CVE-2017-16775"
-]
+{
+	"items": [{
+			"cve_id": "CVE-2019-12310"
+		},
+		{
+			"cve_id": "CVE-2017-16775"
+		}
+	],
+	"expires_at": 1573254000   <------------ seconds since epoch, optional attribute
+}
 ```
 
 ##### On Failure: Authentication Required
@@ -106,11 +117,16 @@ PUT /api/system/vuln_whitelist
 Authorization: xxxxxxx
 Content-Type: application/json
 
-[
-    "CVE-2018-20798",
-    "CVE-2019-0540",
-    "CVE-2019-0590"
-]
+{
+	"items": [{
+			"cve_id": "CVE-2019-12310"
+		},
+		{
+			"cve_id": "CVE-2017-16775"
+		}
+	],
+	"expires_at": 1573254000  <---------  optional
+}
 
 ```
 
@@ -129,8 +145,6 @@ This error happens when the request does not has sufficient permission to do the
 {"code":403,"message":"Forbidden"}
 ```
 
-The value of the system level whitelist will be stored in `properties` table, as there will be only one item.
- 
 ### Add whitelist information to project metadata
 
 Given the way we handle project level configuration currently, we don't plan to introduce new APIs.  Instead, when issuing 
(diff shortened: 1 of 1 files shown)


proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+PUT /api/system/vuln_whitelist
+Authorization: xxxxxxx
+Content-Type: application/json
+
+[
+    "CVE-2018-20798",
+    "CVE-2019-0540",
+    "CVE-2019-0590"
+]
+
```
alexvmw: maybe a whitelist for all distros and then by individual distros? something like this
all:
            "CVE-2018-20798",
	    "CVE-2019-0540",
ubuntu: 
(including different versions)
           "CVE-2018-aaaa",
	    "CVE-2019-bbbb",
alpine:
           "CVE-2018-cccc",
	   "CVE-2019-dddd",

reasonerjt: 1. In Harbor we don't have information about the distro.
2. I don't think it's a good idea split the CVE based on distro, if a system admin reviewed and think a particular CVE can be ignored they should be applied to all distro.  However, I do think it would be helpful to let admin define a pattern of tags that the white list should be applied to, but this would be a future enhancement after v1.9?

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+It proposes to add settings in system/project level to define a white list, so that the vulnerabilities in this list will
+be ignored when Harbor is calculating the vulnerability level of an image.
+
+## Background
+
+Currently Harbor supports "Deployment Security" policy, which prevents images with vulnerabilities higher than threshold from 
+being pulled.  There will be problem if a vulnerability is found and there's no fix for it, in such case, user will have to 
+wait for the fix or disable the policy skip all vulnerabilities.  For some users, they are willing to take the risk after reviewing
+the vulnerabilities and explicitly ignore some of them during the policy checking.
+
```
alexvmw: Based on the background described, it seems if user still wanted to pull image despite CVEs of some criticality found, he needs to put all CVEs (of that criticality) into whitelist, so I think can just remove that 'deployment security' policy.  Also without an expiration date, once this is entered into whitelist, this will persist indefinitely.   But user tolerance for a particular CVE can change over time.  But there are definitely cases where CVE whitelist on top of Deployment Security policy will be useful.

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+* #### System admin defines system level whitelist
+
+There will be only one system level vulnerable white list, which by default is empty.
+The system admin should be able to add or remove CVEs to the whitelist via UI or API.
+
+* #### Project admin reuse the system level whitelist or define the custom whitelist for the project
+
+The "Vulnerability whitelist" will be added to "Deployment Security" as a setting.  A project admin can choose to reuse the 
+system level whitelist or set the custom whitelist that is only effective for the project.
+
```
alexvmw: Project level whitelist settings should override system but this action should only be allowed to be carried out by System admin.  In general, as you go down in granularity (instance -> proj -> repo) settings should have increasing priority, but proj admins should not have ability to easily override policies set by sysadmin

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+##### On Success: OK
+
+```HTTP
+200 OK
+Content-Length: <length>
+Content-Type: application/json
+[
+    "CVE-2019-12310",
```
steven-zou: Maybe adding scope selector to the CVE in the list is more powerful and useful, e.g : `CVE-2019-12310 target tags with [*testing*]`

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+## Proposal
+
+We propose to introduce the "whitelist" of vulnerabilities to project policy to solve this problem.  The system or project admin
+can define a list of CVE, when checking the scan result to compare the the threshold setting in "Deployment Security" policy, the 
+vulnerabilities in the white list will be ignored, such that user can pull an image with certain vulnerabilities higher than the 
+threshold without having to update or disable the policy.
+
+**NOTE:**
+For v1.9.0 there will be a limitation of the number of CVEs in the whitelist, which is 40.
+
```
alexvmw: why 40?

reasonerjt: The 40 items limitation is due to an implementation details, that I wanna reuse the "properties" table in DB and the column is 1024 varchar.

Do you expect the length exceeds 40 in near future?

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+## Proposal
+
+We propose to introduce the "whitelist" of vulnerabilities to project policy to solve this problem.  The system or project admin
+can define a list of CVE, when checking the scan result to compare the the threshold setting in "Deployment Security" policy, the 
+vulnerabilities in the white list will be ignored, such that user can pull an image with certain vulnerabilities higher than the 
+threshold without having to update or disable the policy.
+
+**NOTE:**
+For v1.9.0 there will be a limitation of the number of CVEs in the whitelist, which is 40.
```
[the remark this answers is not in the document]
reasonerjt: It's just to save the implementation effort, if there's strong expectation that we should support longer list, that's do-able.  But I still think we should set a limitation.

proposals/CVE_Whitelist.md
reasonerjt: @alexvmw are you suggesting that only system admin can update the whitelist?

proposals/CVE_Whitelist.md
reasonerjt: It's a valid scenario, but the implementation of scope can vary (based on tag? repo? label? how system level scope and project level scope overwrites?), this will require more effort in design and discussion in community.

We may put that into future enhancement.  But for v1.9 I think we should roll out the feature first and hear the feedback from users and make enhancements.

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+
+In detail, it will involve the following use cases:
+
+* #### System admin defines system level whitelist
+
+There will be only one system level vulnerable white list, which by default is empty.
+The system admin should be able to add or remove CVEs to the whitelist via UI or API.
+
```
alexvmw: harbor web console UI should have an 'upload whitelist' feature

alexvmw: also is there a 'whitelist' view that show all whitelists configured across a harbor instance?

proposals/CVE_Whitelist.md
```diff
@@ -0,0 +1,172 @@
...
+being pulled.  There will be problem if a vulnerability is found and there's no fix for it, in such case, user will have to 
+wait for the fix or disable the policy skip all vulnerabilities.  For some users, they are willing to take the risk after reviewing
+the vulnerabilities and explicitly ignore some of them during the policy checking.
+
+## Proposal
+
+We propose to introduce the "whitelist" of vulnerabilities to project policy to solve this problem.  The system or project admin
+can define a list of CVE, when checking the scan result to compare the the threshold setting in "Deployment Security" policy, the 
+vulnerabilities in the white list will be ignored, such that user can pull an image with certain vulnerabilities higher than the 
+threshold without having to update or disable the policy.
```
alexvmw: is Webhook still triggered for whitelisted CVEs?

alexvmw: Also please make sure pull log for an image with whitelisted CVEs clearly show that image was pulled despite CVEs found.

proposals/CVE_Whitelist.md
reasonerjt: If we treat whitelist as project level attributes or project level policy, we need a systematic design for overview for project level policies.

proposals/CVE_Whitelist.md
alexvmw: @reasonerjt if project admin has full control of project settings (which overrides system, this part is correct at least) then there would be issues, and not just for this.  I think small granularity should have highest priority but privileges is a different issue.  Otherwise you can project admin can always override system admin actions.

proposals/CVE_Whitelist.md
alexvmw: what happens when you find more than 40 CVEs of some critically which blocks your pull? how do you get around that?

proposals/CVE_Whitelist.md
alexvmw: @reasonerjt OK that's fine

proposals/CVE_Whitelist.md
reasonerjt: Agreed, at least from storage or API layer, we should not have such limitation.

proposals/CVE_Whitelist.md
reasonerjt: I assume by pull log you mean the text log message?  That will be done for sure.

proposals/CVE_Whitelist.md
reasonerjt: I kind of agree with your opinion that we should remove/disable the policy if user wanna pull a vulnerable image.
But this whitelist was required by a few users and we've made the decision we wanna implement it in v1.9.

I also think it's a good suggestion to add expiration time, I'm not sure if we should put it as P0, but I'll update the proposal to make the data structure of API and DB schema more extensible, to allow we add more attributes to each white list in the future.

proposals/CVE_Whitelist.md
reasonerjt: I think whether we allow project admin to overwrite system level white list at project level is a matter of decision.

But personally I think project admin can have full control for objects under project scope, so such "overwriting" should be allowed.  If you think such overwriting is harmful, we may better define system setting and project setting, instead of letting them set same things.  
Additionally if system admin is not happy with what project admin does he can undo that at project level, or even assign another project admin, as for now system admin is the `root` of whole Harbor.

alexvmw: Actually I agree. The system setting is to reduce repetitive work, to cover everything with one action.  Project admin overwriting it is fine.

proposals/CVE_Whitelist.md
alexvmw: great, as long as we are aware of the issue.

proposals/CVE_Whitelist.md
alexvmw: Yep a simple text log will be fine.  ditto on the webhook

proposals/CVE_Whitelist.md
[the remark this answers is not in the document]
reasonerjt: Resolving as the limitation is removed.

PR #88, opened 2019-06-20
propose to support external syslog endpoint

ninjadq:
propose to add external syslog endpoint
so harbor's log can send to outer log service


proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+template(name="DynaFile" type="string" string="/var/log/docker/%programname%.log")
+
+if $programname != "rsyslogd" then {
+{% if endpoint == "local"}
+        action(type="omfile" dynaFile="DynaFile")
+{% else %}
+        action(type="omfwd" Target="10.192.233.234" Port="5140" Protocol="tcp" Template="RSYSLOG_SyslogProtocol23Format")
+{% end %}
+}
+```
```
heww: How to deal with this config in harbor helm chart ?

proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+log:
+  level: info
+  rotate_count: 50
+  rotate_size: 200M
+  # location: /var/log/harbor
+  external_endpoint:
+    protocol: tcp
+    host: localhost
+    port: 5140
+  ```
```
heww: Can rotation work in syslog forward type? If not, maybe make rotate_count, rotate_size and location together will be better. 

What about the following config ?
```yaml
log:
  level: info
  type: local # local or external
  local:
    rotate_count: 50
    rotate_size: 200M
    location: /var/log/harbor
  external:
     target: xxx.xxx.xxx.xxx
     port: xxxx
     protocol: tcp   
```

proposals/new/support_external_syslog_endpoint.md
ninjadq: we don't need to handle the log when running harbor in k8s. k8s will process logs by user's config

proposals/new/support_external_syslog_endpoint.md
ninjadq: The logic in this proposal is that if you uncomment the external_endpoint then forward log to an external endpoint or by default using the local file. 

Actually, log rotate is provided by logroate and cron job. So these two configurations you mention are not targeted to the same subject. Although the log files are empty if you export the logs to outside endpoint

proposals/new/support_external_syslog_endpoint.md
heww: ok, i found that there is no `harbor-log` container in helm chart.

proposals/new/support_external_syslog_endpoint.md
heww: Do you mean that rotation is not provided in `harbor-log` container ?

ninjadq: yes, in `harbor-log` container, there are two services, `rsyslogd` and `logretate`, the config items set in log section will render each service's real config file

proposals/new/support_external_syslog_endpoint.md
ninjadq: yes, `harbor-log` only existed in the standalone harbor

proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Implementation
+
+1. Add external log endpoint related options in harbor.yml
+
+```yaml
+# Log configurations
+log:
+  level: info
+  rotate_count: 50
```
reasonerjt: Will the rotate be applied to external logging endpoint?

proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+```conf
+#  Rsyslog configuration file for docker.
+
+template(name="DynaFile" type="string" string="/var/log/docker/%programname%.log")
+
+if $programname != "rsyslogd" then {
+{% if endpoint == "local"}
+        action(type="omfile" dynaFile="DynaFile")
+{% else %}
+        action(type="omfwd" Target="10.192.233.234" Port="5140" Protocol="tcp" Template="RSYSLOG_SyslogProtocol23Format")
```
reasonerjt: what happens if the external logging endpoint is down?

proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+1. Add external log endpoint related options in harbor.yml
+
+```yaml
+# Log configurations
+log:
+  level: info
+  rotate_count: 50
+  rotate_size: 200M
+  # location: /var/log/harbor
+  external_endpoint:
```
reasonerjt: We should be able to support multiple endpoints?

proposals/new/support_external_syslog_endpoint.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Background
+
+Currently, Harbor stores the log files in local. However most modern application especially the Cloud-Native apps will send their logs to remote host. In that way users do not need connect the machine to check the logs anymore, and they can access a central log service to analyse the logs which often provides user friendly interface and more powerful analysis tools. To support this Harbor should provide a way to output its logs to external endpoint.
+
+Moreover Harbor already has a component harbor-log in which a rsyslog server is running to collect logs generated by other containers and store them to local machine.
+
+## Proposal
+
+1. Add the ability to harbor-log container to forwarding collected logs to other source
```
reasonerjt: Why we choose to use harbor-log to forward the log instead of using docker's logdriver?

proposals/new/support_external_syslog_endpoint.md
ninjadq: I don't think this is the scope of Harbor. There are so many third parties' solutions can do this. like `fluentd`, `fluentbit`, `logstash`.

proposals/new/support_external_syslog_endpoint.md
ninjadq: if logging endpoint is down, the logs generated in this period will be lost

proposals/new/support_external_syslog_endpoint.md
ninjadq: the config related to rotate is not for rsyslog. The logrotate will read these configs.

proposals/new/support_external_syslog_endpoint.md
ninjadq: because the log format of harbor components are varied. we need to format them in a uniform style. log container can do this kind of thing

proposals/new/support_external_syslog_endpoint.md
heww: Your reply means that rotation is provided in `harbor-log` container, `harbor-log` run two service, one of it implement the rotation and only work for the local file logs. 

I suggest to make rotate_count, rotate_size and location together.

proposals/new/support_external_syslog_endpoint.md
reasonerjt: so this should be applied to local only.

proposals/new/support_external_syslog_endpoint.md
reasonerjt: Think the design of log framework or even rsyslog, a log can have multiple handlers, I think we should also follow this pattern.

i.e. 
1) allow local and external log endpoint work at the same time
2) allow multiple external log endpoints work at the same tie.

PR #89, opened 2019-07-01
add proposal patch for the tag retention feature

steven-zou:

This proposal is a patch of the original [proposal](https://github.com/goharbor/community/blob/master/proposals/5882-Tag-Retention-Policies.md) about tag retention policy feature to revise some technical designs which reflect the latest related discussions.


proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+  } `json:"rule_id"`
+  Artifact string `json:"tag"` // full path: :ns/:repo:tag
+  Timestamp time.Time string `timestamp`
+  Operation string `json:"operation"`
+}
+```
+
+**Notes:** Some helper methods of the models may not listed in this proposal.
+
+##### Manager Interface
```
nlowe: There are a ton of interfaces here, some of them feel like they could just be vanilla types instead. I just want to make sure we're not over-complicating things given the shorter time table.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+
+#### Overall
+
+The following graph shows the overall architecture of tag retention.
+![arch](./images/retention/arch.png)
+
+* API controller provides unified interface at the top level and serve the request from the upper rest API layer.
+* Manager implements the CRUD operations of policy, rule templates and execution history based on the DAOs.
+* Scheduler schedules/unschedules a periodic job to generate interval events which launch the retention job.
+* The hook handler will listen to the retention job execution status hook events and update the related stats info and history by talking to the manager, or listen to the schedule events and tell the launcher to launch retentions once events coming in.
```
nlowe: Unless this hook handler is already a component in Harbor it feels like an unnecessary component. Is there a reason the API can't talk to the launcher / retention manger directly?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+The relationship among rules in one policy will be **OR** way, e.g:
+
+Policy attached to the project _hello-world_ has two rules,
+
+Rule-0 outputs keep `[repo_x/tag1, repo_y/tag2]`
+Rule-1 outputs keep `[repo_x/tag3, repo_z/tag5]`
+
+after execution, the tags `[repo_x/tag1, repo_y/tag2, repo_x/tag3, repo_z/tag5]` will be kept, all others will be cleared.
+
+**Notes:** Although _AND_ relationship is not covered in this proposal, the framework proposed here will consider enough flexibilities for the future _AND_ supporting.
```
alexvmw: 'AND' needs ability for users to add priority to rules

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+
+### User Stories
+
+* As a (system/project) admin, Steven can view and remove the rules of the retention policy attached to the specified project.
+* As a (system/project) admin, Steven can add new rule into the policy attached with the specified project by selecting the rule template and providing the required parameter.
+* As a (system/project) admin, Steven can select pre-defined schedule (daily, weekly or monthly etc.) or set customized CRON like (* * 24 * * *) schedule to launch the retention process.
+* As a (system/project) admin, Steven can luanch a `dry run` process and check the overall summary containing what tags will be cleared if launching this policy.
+* As a (system/project) admin, Steven can run the retention policy immediately (manually trigger).
+* As a (system/project) admin, Steven can check the retention executing history which includes overall status, related timestamps and what tags are cleared.
+* As a (system/project) admin, Steven can disable the retention by setting `None` schedule or clearing the policy.
```
alexvmw: what happens when you disable an ongoing retention policy where things are being deleted ?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+## Proposal
+
+### Overall Design
+
+#### Retention Policy
+
+The request of performing retention will be defined as a retention policy. That means retention will be policy based. A retention policy should include:
+
+* a collection of related retention rules defined in the below section.
+  * option of limiting the number of rules in a policy is supported
```
alexvmw: is this limit set by harbor admin in web console or in backend somehow?  does changing the value require restarting the harbor service?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+
+| Rule Template |  Parameter |  Selector |  Priority |
+|---------------|------------|-----------|-----------|
+|Keep everything|  NO        | **        |  P0       |
+|keep most recent X tags | number of tags| All  | P0       |
+|Keep no older than X days | life time of tag | All | P0   |
+|Keep X tags    | number of tags|  All   |  P1       |
+|Keep most recent downloaded X tags | number of tags | All | P1 |
+
+### User Stories
```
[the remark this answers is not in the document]
steven-zou: Added

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+  // Which scope the policy will be applied to
+  Scope *Scope `json:"scope"`
+  // Exceptional repositories
+  Exclusions []*Exclusion `json:"exclusions"`
+}
+
+// Capacity declares max number of rules in a policy
+type Capacity struct {
+  Cap int `json:"cap"`
+  Occupied int `json:"occupied"`
```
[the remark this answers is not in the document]
steven-zou: Leave flexibilities for the future if the settings are exposed as configurable options.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+
+##### Model for policy and rule
+
+```go
+package retention
+
+// Policy object
+type Policy struct {
+  // UUID of the policy
+  ID string `json:"id"`
```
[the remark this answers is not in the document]
steven-zou: Yes, for a new implementation, I'd like to use UUID not database row id number to identify the data.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+type Trigger struct {
+  // Const string to declare the trigger type
+  // 'Schedule'
+  Kind string `json:"kind"`
+  // Settings for the specified trigger
+  // '[cron]="* 22 11 * * *"' for the 'Schedule'
+  Settings map[string]interface{} `json:"settings"`
+  // References of the trigger
+  // e.g: schedule job ID
+  References map[string]interface{} `json:"references"`
```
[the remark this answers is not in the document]
steven-zou: `Settings` keep the required parameters or configuration options for the specified trigger. These settings are configured by the users. For example, if the trigger is `schedule`, the settings will be cron pattern like `* 2 10 * * *`.

Reference keep more information generated by the system. The same example, if the trigger is `schedule`, the system will generate a periodic job to run the loop. As we need the job ID to control the job, this ID will be kept into the reference part.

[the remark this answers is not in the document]
steven-zou: Just for clear differentiation

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+
+// Rule of retention
+type Rule struct {
+  // UUID of rule
+  ID string `json:"id"`
+  // Priority of rule when doing calculating
+  // Placeholder for extending
+  Priority int `json:"priority"`
+  // Template ID
+  Template string `json:"template"`
```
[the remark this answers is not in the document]
steven-zou: It does not cache the implementation, cache the implementation class reference which can be used to build new instance dynamically. Some example code here:

```go
type Evaluator interface {
  Process(...) error
}

type RuleFactory func(options ...interface{}) Evaluator

type MostRecentXRule struct {
  x int
}

func (mrx *MostRecentXRule) Process(...) error {
  return nil
}

func MostRecentXFactory (x int) Evaluator {
  return &MostRecentXRule{x}
}

var RuleRegistry = make(map[string]RuleFactory)

func init(){
  RuleRegistry["mostRecentX"] = MostRecentXFactory
}

// When using 

var r Rule

//...
if factory, ok := RuleRegistry[r.TemplateID]; ok {
  inst := factory(r.Parameters["x"].Int())
  if err := inst.Process ; err!=nil{
   returtn err
 }
}


```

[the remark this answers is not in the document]
steven-zou: Another way is using the `reflect` package

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,645 @@
...
+}
+```
+
+##### Scheduler Interface
+
+```go
+package policy
+
+// Scheduler of launching retention jobs
+type Scheduler interface {
```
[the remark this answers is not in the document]
steven-zou: They're all using the periodic job provided by the job service which will be ok under HA mode

proposals/Tag-Retention-Patch.md
steven-zou: Agree, just define what we need. However, interface-based programming will make the testing and easier and structure of service more clear (easy to do modification/refactor in the future).

proposals/Tag-Retention-Patch.md
steven-zou: Yes, they can. This hook handler is defined to provide logic to handle the events specified to the retention service. It can not share with others. Moreover, it's different from the hook/handlers things related to the implementing Harbor webhook.

The logic can be directly put into the API, but it's more clear to wrap it into a separate interface. Same logic, just different code style.

proposals/Tag-Retention-Patch.md
steven-zou: Yes

proposals/Tag-Retention-Patch.md
steven-zou: Retention processes of some repositories will be terminated if they're still not completed.

proposals/Tag-Retention-Patch.md
steven-zou: So far, it's hardcoded in the plan.

proposals/Tag-Retention-Patch.md
steven-zou: Move hook handler logic to the API controller is acceptable.

proposals/Tag-Retention-Patch.md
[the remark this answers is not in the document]
steven-zou: Removed the capacity. It can be checked in other ways.

proposals/Tag-Retention-Patch.md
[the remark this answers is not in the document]
nlowe: I see, so it's a factory mapping. Thanks for the example, that was super helpful.

I think I took the opposite approach with the initial crack I took at the first proposal:

https://github.com/goharbor/harbor/blob/aeebcc29518cdb382122641dc697c4c02dffe202/src/retention/chain/builder.go#L24-L54

Essentially each filter metadata has a `type` property (roughly analogous to the `TemplateID` in this patch), but I left the "factory" functions as public functions exposed on the types directly. Functionally, these do the same thing I think, so I'm fine with whatever route we want to go with, the updated proposal just feels like a lot of extra boilerplate.

nlowe:
When this gets merged we should also circle back on goharbor/harbor#6654 as well to re-visit what tracking issues need created.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,639 @@
...
+In this proposal,
+
+* only `Keep` action is supported to purchase the retention goal with a safe and positive way. All the resources NOT in the `Keep` list will be cleared.
+* the following decorated conditions will be prioritized and covered in order:
+  * `most recent` which is evaluated based on the _push time_ of the resource
+    * parameter required: the number of resources (see above example 1)
+  * `no older than` which is evaluated based on the _push time_ of the resource
+    * parameter required: the life time with day unit (see above example 2)
+  * `most recent downloaded` which is evaluated based on the _pull time_ of the resource
+    * parameter required: the number of resources (see above example 3)
```
alexvmw: can we consider both times, ie most 5 recent images (regardless of whether push or pull time), so looking at both activities.

bitsf:
what about db model design below?

type RetentionPolicy struct {
	ID int64
	// 'system', 'project' and 'repository'
	ScopeLevel string
	ScopeReference int64
	TriggerKind string
	TriggerSettings string
	// json format, include algorithm, rules, exclusions
	Data string
	CreateTime time.Time
	UpdateTime time.Time
}

type RetentionExecution struct {
	ID int64
	PolicyID int 64
	Status string
	StatusText string
	Dry bool
	// manual, scheduled
	Trigger string
	Total int
	Succeed int
	Failed int
	InProgress int
	Stopped int
	StartTime time.Time
	EndTime time.Time
}

type RetentionTask struct {
	ID int64
	ExecutionID int64
	// image, chart
	ResourceType string
	Resource string
	Status string
	StartTime time.Time
	EndTime time.Time
}

type RetentionScheduleJob struct {
	ID int64
	Status string
	PolicyID int64
	JobID int64
	CreateTime time.Time
	UpdateTime time.Time
}

steven-zou:
UE mockup link:https://www.figma.com/proto/0bSkyzvchcOCrwSWROIivtIb/Harbor-v1.9?node-id=418%3A265&viewport=1636%2C1073%2C0.9472222328186035&scaling=min-zoom

**NOTES:** access permissions may be needed

steven-zou:
@nlowe 

Per your comments and some ongoing work of interface defining, I updated the arch diagram and related interface definitions. Please look over and see if you want to add additions.

Btw, based on the epic and its subsequent tasks you created before and our effort plans, I re-assign the owners of the tasks. Please also take a look and see if you have any other suggestions. Your work is to implement the two P0 rules `Retain latest k tags` and `Retain the last x days tags`.

**NOTES:** 
* The base code has committed to the feature branch `feature/tag_retention` and some core logic has been implemented.
* If needed, we can schedule a zoom call to go through the work

proposals/Tag-Retention-Patch.md
steven-zou: Suggest sticking to only one bar.

steven-zou: push or pull times

proposals/Tag-Retention-Patch.md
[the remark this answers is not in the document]
steven-zou: Yeah, a similar thing. Factory method is easier than using the `reflect` package and also simple for passing parameters.

steven-zou:
> When this gets merged we should also circle back on [goharbor/harbor#6654](https://github.com/goharbor/harbor/issues/6654) as well to re-visit what tracking issues need created.

@nlowe 

owners re-assigned and only add 1 more task for UI implementation. Your tasks cover almost all of the current work. You're awesome!  That also means the work proposed in the original proposal and the work in this patch proposal are very similar.

nlowe:
@steven-zou Yeah, the UI mockup appears to be locked down, that link 404's for me.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,838 @@
...
+  //    []*res.Result : the processed results
+  //    error         : common error object if any errors occurred
+  Process(artifacts []*res.Candidate) ([]*res.Result, error)
+
+  // Add a rule evaluator for the processor
+  //
+  //  Arguments:
+  //    evaluator rule.Evaluator    : a rule evaluator
+  //    selectors []res.Selector    : selectors to narrow down the scope (&& adopted), optional
+  AddEvaluator(evaluator rule.Evaluator, selectors []res.Selector)
```
nlowe: This (and `AddPerformer` below) feels out of place. Don't things like this (that modify the underlying instance that implements the interface) usually belong as arguments to a factory func / ctor?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,838 @@
...
+  //  Arguments:
+  //    artifacts []*res.Candidate : candidates for processing
+  //
+  //  Returns:
+  //    []*res.Candidate : matched candidates for next stage
+  //    error            : common error object if any errors occurred
+  Process(artifacts []*res.Candidate) ([]*res.Candidate, error)
+
+  // Specify what action is performed to the candidates processed by this evaluator
+  Action() string
```
nlowe: From earlier in the proposal:

> In this proposal,
> 
> * only `Retain` action is supported to purchase the retention goal with a safe and positive way. All the resources NOT in the `Retain` list will be cleared.

Isn't this always the same value, or is this for future expansion?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,838 @@
...
+type Processor interface {
+  // Process the artifact candidates
+  //
+  //  Arguments:
+  //    artifacts []*res.Candidate : process the retention candidates
+  //
+  //  Returns:
+  //    []*res.Result : the processed results
+  //    error         : common error object if any errors occurred
+  Process(artifacts []*res.Candidate) ([]*res.Result, error)
```
nlowe: One important thing I also wanted to ask: In the original proposal, filters were guaranteed to receive artifacts sorted by creation date, newest artifacts first. Does that still hold, or do processors need to sort the artifacts if they expect a certain order?

proposals/Tag-Retention-Patch.md
steven-zou: @nlowe 

For future expansion. In the V1.9, we only need to support `retain`.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,838 @@
...
+type Builder interface {
+  // Builds runnable processor
+  //
+  //  Arguments:
+  //    rawPolicy string : the simple retention policy with JSON format
+  //
+  //  Returns:
+  //    Processor : a processor implementation to process the candidates
+  //    error     : common error object if any errors occurred
+  Build(rawPolicy string) (alg.Processor, error)
```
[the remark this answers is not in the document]
steven-zou: @nlowe 

Yes, it's correct.

After check code in js, the job service has recovered the structure. Maybe we do not need passed policy as a raw string.

In the feature branch, I have updated the interface to use a pointer.

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,838 @@
...
+```
+
+##### Launcher Interface
+
+```go
+package retention
+
+// Launcher provided function to launch the jobs to run retentions
+type Launcher interface {
+  Launch(policy *policy.Metadata) ([]string, error)
```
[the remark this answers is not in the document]
steven-zou: @nlowe 

Forgot to add detailed comments for this interface.

It's a list containing the IDs of the launched async jobs.

steven-zou:
@nlowe 

The rule list defined in the proposal is updated. Add two more rules. Is it ok for you to implement the other two ones?

proposals/Tag-Retention-Patch.md
steven-zou: @nlowe 

Good question. In the new proposal, the order of the receiving artifacts is not guaranteed. The rule evaluator needs to cover the order itself. Actually, that will be more flexible. Some case like `most recent active` need to do evaluations based on both _push time_ and _pull time_.

proposals/Tag-Retention-Patch.md
steven-zou: @nlowe 

I totally understand your concern. Actually, the original version is using a factory method to inject the related evaluators. The thought of proposing this version explicitly defines the structure of the processor.  

I'm thinking if we should roll back to the original version.

steven-zou: > @nlowe
> 
> I totally understand your concern. Actually, the original version is using a factory method to inject the related evaluators. The thought of proposing this version explicitly defines the structure of the processor.
> 
> I'm thinking if we should roll back to the original version.

Let's roll back to the version only having `Process` method.

steven-zou: Proposal updated. I'll update the code later.

steven-zou:
@mmpei @kofj @cd1989 

I would be grateful if you guys can review this proposal and give your decision vote before the end of this week. Thanks!

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,814 @@
...
+For empty policy (no any rules), `Retain everything` will be performed as default design.
+
+The rule will be pre defined as full rule template and initialized as rule instance by letting admins provide the related required parameters when adding to its parent policy.
+
+In summary, the following rule templates will be supported in priority order one by one:
+
+| Rule Template |  Parameter |  Selector |  Priority |
+|---------------|------------|-----------|-----------|
+|Retain everything|  NO        | **        |  P0       |
+|Retain most recently pushed # images | number of tags| All  | P0 |
```
cd1989: `creation time`  of images not considered ?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,814 @@
...
+* a collection of related retention rules defined in the below section.
+  * option of limiting the number of rules in a policy is supported
+* a trigger approach declaring how to launch the retention operation process
+  * CRON-like schedule
+  * Pre-defined `None` represents no schedule enabled
+* a scope declaration about the target range to apply the retention operation (each scope only needs attach 1 policy as a default design)
+  * system level
+  * project level
+  * repository level
+* an exclusion list including the exceptional repositories which policy will not be applied to
```
cd1989: To exclude projects, it's `<project>/**` ?

proposals/Tag-Retention-Patch.md
cd1989: Why should we set this limit ?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,814 @@
...
+
+See original background [statements](https://github.com/goharbor/community/blob/master/proposals/5882-Tag-Retention-Policies.md#background).
+
+## Proposal
+
+### Overall Design
+
+#### Retention Policy
+
+The request of performing retention will be defined as a retention policy. That means retention will be policy based. A retention policy should include:
```
cd1989: If multiple policy created for the same project, will it be difficult to understand what will happen actually to the project images.  To me, I care what's the final effect to a project or a repo.

Why not consider retention policy as configure of a project, that's to say, one policy to one project. So that we can manage the retention for a project easily.

proposals/Tag-Retention-Patch.md
nlowe: I think that makes sense. Can you add a note to the patch so that's clear, especially because this is a change from the original proposal?

proposals/Tag-Retention-Patch.md
steven-zou: @nlowe 

Code updated in PR https://github.com/goharbor/harbor/pull/8275

proposals/Tag-Retention-Patch.md
steven-zou: Sure, will add notes

proposals/Tag-Retention-Patch.md
steven-zou: Has use cases? @cd1989

proposals/Tag-Retention-Patch.md
steven-zou: ignore it. It has been removed already. Forgot to remove it from the doc @cd1989

proposals/Tag-Retention-Patch.md
steven-zou: For the end user, the case will be one project one policy (with multiple rules). That's the design.

@cd1989

cd1989: From the API `POST /api/retentions`, policy is not bounded to a project, that means users can create different policies for a project, am I right ?

proposals/Tag-Retention-Patch.md
cd1989: For base images like `ubuntu`, `alpine`, there is no guarantee for the order to push different versions. If users want to retain the newest versions, by using `creation time` would be more precise.

proposals/Tag-Retention-Patch.md
steven-zou: @cd1989 

The API is designed for retention policy management. For the relationship between the project and its corresponding retention policy (rules), it will be saved in the project metadata.

In the project metadata, there will be a property like 'retention' to point to the policy. Each policy can only point to one. And no way changed in the console. 

It's a loose couple binding way.

proposals/Tag-Retention-Patch.md
steven-zou: @alexvmw 

Any suggestions?

steven-zou:
> @steven-zou Yeah, the UI mockup appears to be locked down, that link 404's for me.

Let me paste the picture here

steven-zou:
# Main page
Screen Shot 2019-07-11 at 10 20 16

# Rule adding dialog

Screen Shot 2019-07-11 at 10 20 30

# Running retention

Screen Shot 2019-07-11 at 10 20 51

proposals/Tag-Retention-Patch.md
xaleeks: I think it's just semantics, in the UI/UX officially, we are calling each configured policy (capped at 15) a 'retention rule', we settled on the word rule.  You can interpret the combination of all configured rules (run stacked on top of each other and taking the union of the captured 'to retain' image sets) as a retention policy for that project, but the word 'policy' does not appear on the web console @cd1989

proposals/Tag-Retention-Patch.md
xaleeks: I'm not sure I understand 'For base images like ubuntu, alpine, there is no guarantee for the order to push different versions'.  Recent is defined by the push time of a completed push action, but it could be something else.  @cd1989

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,851 @@
...
+
+* a collection of related retention rules defined in the below section.
+  * option of limiting the number of rules in a policy is supported
+* a trigger approach declaring how to launch the retention operation process
+  * CRON-like schedule
+  * Pre-defined `None` represents no schedule enabled
+* a scope declaration about the target range to apply the retention operation (each scope only needs attach 1 policy as a default design)
+  * system level
+  * project level
+  * repository level
```
mmpei: how to deal with the confliction between project level and repository level? even with system level?
i think repository level is deficult to manage. and could this be replaced by regular expression 'reponame/tag*' in project level?

proposals/Tag-Retention-Patch.md
```diff
@@ -0,0 +1,851 @@
...
+> Retain the most recent active 5 images without labels [x,y] in repositories a,c
+
+The retention rule can be abstracted to a general format with the following segments:
+
+`RULE=[ACTION][DECORATED CONDITION(PARAMETERS)][TAG SELECTORS][SCOPE SELECTORS]`
+
+In this proposal,
+
+* only `Retain` action is supported to purchase the retention goal with a safe and positive way. All the resources NOT in the `Retain` list will be cleared.
+* the following decorated conditions will be prioritized and covered in order:
```
mmpei: i don't understand that why rules have a priority if they are combined by 'or'.

mmpei:
another thing that makes me confused is whether the new pushed images in a new repository will be removed if i defined a policy that retains images in some other repositories in project level?

proposals/Tag-Retention-Patch.md
steven-zou: Leave flexibility for the `and` support in the future

proposals/Tag-Retention-Patch.md
steven-zou: @mmpei 

In the project level, we do support regexp like the pattern you mentioned 'reponame/tag*.

In the future, if we support different levels at the same time, the calculating algorithm maybe something like `inherited` and then bunch them as a big rule group to complete the evaluation process.

steven-zou:
> another thing that makes me confused is whether the new pushed images to a new repository will be removed if i defined a policy that retains images in some other repositories in project level?

@nlowe 

The deleting action only removes the corresponding ref, no any conflicts/influences to the other repos in other projects.

proposals/Tag-Retention-Patch.md
mmpei: ok

proposals/Tag-Retention-Patch.md
cd1989: Let me make it clear. For an image, suppose there are v1.0, v2.0 tags, if users want to retain new v1.0. But v2.0 is pushed before v1.0, then by checking `push time`, v2.0 will be removed.

proposals/Tag-Retention-Patch.md
cd1989: API semantics also counts.

steven-zou:
As this PR is approved by @nlowe , @mmpei , @kofj , @cd1989 and other maintainers, I'll merge it now. Thank you all!

PR #94, opened 2019-07-17
Add other P2P wg members

cd1989:
Signed-off-by: cd1989 <chende@caicloud.io>


steven-zou:
Let's merge this update

PR #98, opened 2019-07-31
Proposal: Pluggable Image Vulnerability Scanning

danielpacak:
This is a combination of #82 and #90 and [PRD by Alex Xu](https://docs.google.com/document/d/1CGqhf1KNzajmg69h8K6OoFhx7mLXSbGjukZGCXM82vs/edit).

The main file to read is `proposals/pluggable-image-vulnerability-scanning_proposal.md`.


proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+The main focus of this proposal and the subsequent sections is the Scanner Adapter API design as it must be stable for
+Scanner Adapter implementers. Harbor-internal design is secondary as this can evolve independently of the Scanner
+Adapter API as long as the contract between the two is set.
+
+### Image Scanning with Scanner Adapter
+
+The indirection introduced by the Scanner Adapter framework allows implementation of a generic ScannerJob which in turn
+uses a generic Scanner Adapter Client. Both are agnostic to the underlying scanner internals or upgrades.
+
+Such architecture is also extensible to triggering multiple Scanner Adapters for a single scan request. This would require
```
reasonerjt: In the Diagram below, take Clair as example, I assume the `Scan adapter` and `Clair adapter` and `ClairAPI` should be run in one process?  If the answer is yes I don't see the need to add another layer of adapter, considering the Clair and Aqua have to be run in different processes.

I also recall that we've made decision to let `scan controller` and `scan job` coordinate the whole scan process in future if we want to support aggregation of scan result.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+aggregating ScanReports coming from each Scanner Adapter but is certainly doable without changing the contract
+between Harbor service and Scanner Adapters. For example, there might be a `ScanReportAggregate` struct which holds
+a collection of `ScanReport`s. Such an aggregate can be visualised in Harbor Web console as a tabbed pane, with each tab
+corresponding to a scanner involved.
+
+![harbor-scanner-adapter-deployment][harbor-scanner-adapter-deployment-img]
```
reasonerjt: In the deployment diagram why is there a `Read` action between `scan adapter` and `registry`?  I believe it's the scanner that actual pulls image from registry?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+7. At this state the scanner can pull the image and start scanning and generating vulnerability reports.
+8. The ScannerJob blocks to pull the ScanReport from Scanner Adapter. Note the identifier of ScanRequest and the mime
+   type passed to retrieve recognizable ScanReport.
+9. The ScannerJob transforms ScanReport to the ComponentsOverview model.
+10. The ScannerJob saves the ComponentsOverview to the Harbor DB along with the identifier of the ScanRequest and the
+   mime type of ScanReport. (The identifier of a ScanRequest plays role of DetailsKey as is currently the case for
+   Clair integration.)
+11. The User requests refresh of the vulnerabilities report for a given image.
+12. The Harbor API retrieves the DetailsKey for the image Digest.
+13. The Harbor API pulls ScanReport from Scanner Adapter.
```
reasonerjt: As for item 9,10,13, I recall @danielpacak  suggested that the scan result/report has to be stored in Harbor's DB?  If so, Harbor API should be able to pull the result from DB instead of from the `Scanner Adapter`?

Additionally, if we have the scan result from in the DB the scannerJob may not need to transform it to `ComponentsOverview` model.

proposals/pluggable-image-vulnerability-scanning_proposal.md
lizrice: My understanding of this is that each adapter is an abstraction, mapping requests from the common ScannerAdapter APi to each individual scanner's own API. Maybe this would be clearer if Clair wasn't the only scanner with a separate box for its API in the diagram?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+         per-scanner UI work
+      3. Simpler for non-admins to understand results without detailed knowledge of scanners
+   2. Disadvantages
+      1. Lowest common denominator for scanners
+      2. Harbor limited to consuming only things it understands, so must keep up with scanners to add new capabilities
+
+## Compatibility
+
+1. Scanner Adapter API must be versioned
+2. Preserving previous scan results?
```
lizrice: Unless the current Clair implementation preserves previous scan results, I'd suggest this should be out of scope for the first version of pluggable scanner support

proposals/pluggable-image-vulnerability-scanning_proposal.md
lizrice: It looks to me like this is intended to be back-compatible with the current Clair approach

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+Harbor during initiation of a scan.
+
+### Scanner Adapter API
+
+Scanner vendors are supposed to implement Scanner Adapters exposing the Scanner Adapter API as specified in
+[Scanner Adapter OpenAPI Specification v1.0.0](assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml).
+
+> Note: OpenAPI spec yaml file can be opened in the online [Swagger Editor](https://editor.swagger.io).
+
+- The deployment method is up to the vendor as long as the mounted API endpoint ULR is accessible to Harbor services.
```
lizrice: typo URL

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+13. The Harbor API pulls ScanReport from Scanner Adapter.
+ 
+### Artifact Data Access
+
+Scanners need access to the artifact data, images in this case. Scanners will retrieve data from Harbor for analysis
+using existing Harbor APIs and optionally credentials. The specific APIs used will vary by the artifact being analyzed,
+but for the initial image analysis, the image data is retrieved by the scanner using the Docker Registry v2 API exposed
+by Harbor. Necessary credentials are presented to the scanner by the scanner adapter as provided in the API call by
+Harbor during initiation of a scan.
+
```
reasonerjt: Considering Harbor has policy to block pulling vulnerable images or unsigned images, there has to be some contract between Harbor and scanner that when the incoming request containing certain header, such policies should be disabled.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+- The deployment method is up to the vendor as long as the mounted API endpoint ULR is accessible to Harbor services.
+- ScannerJob UUID should be a scan request ID used in conjunction with the artifact ID, i.e. artifact repository and
+  digest, and not treated as persisted in the Scanner Adapter.
+- The lifetime of ScannerJob UUIDs in a Scanner Adapter is defined by the adapter. A ScannerJob may get a 404 response
+  status for a UUID and should treat it as failed and return a failure in the job. Harbor is expected to execute a new
+  ScannerJob in that case.
+- Scanner Adapters are not expected to persist scan reports or make responses for ScanJob UUID + artifact ID immutable.
+- ScannerJob UUIDs are correlation IDs, not result set identifiers.
+- Scanner Adapter API leverages content negotiation by using mime-types in `Accept` header to define schema of a result
```
reasonerjt: the `/scan/{scan_request_id}/report` is not consistent with that was defined in `scanner-adapter-openapi-v1.0.0.yaml` where it is defined as `/scan/{scan_request_id}`

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+> Note: OpenAPI spec yaml file can be opened in the online [Swagger Editor](https://editor.swagger.io).
+
+- The deployment method is up to the vendor as long as the mounted API endpoint ULR is accessible to Harbor services.
+- ScannerJob UUID should be a scan request ID used in conjunction with the artifact ID, i.e. artifact repository and
+  digest, and not treated as persisted in the Scanner Adapter.
+- The lifetime of ScannerJob UUIDs in a Scanner Adapter is defined by the adapter. A ScannerJob may get a 404 response
+  status for a UUID and should treat it as failed and return a failure in the job. Harbor is expected to execute a new
+  ScannerJob in that case.
+- Scanner Adapters are not expected to persist scan reports or make responses for ScanJob UUID + artifact ID immutable.
+- ScannerJob UUIDs are correlation IDs, not result set identifiers.
```
reasonerjt: Does it imply the call should generate the uuid?
What's the benefit of this design comparing to let the scan-adapter generates the UUID?

If we look at the adapter as a service, it would be confusing if the service receives two concurrent design with the same "UUID" from two callers.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+   ```
+   SCAN_REQUEST_ID=$(uuidgen | tr "[:upper:]" "[:lower:]")
+   ```
+3. Submit the scan request:
+   ```
+   curl http://scanner-adapter:8080/api/v1/scan \
+   -H 'Content-Type: application/json; charset=utf-8' \
+   -d @- << EOF
+   {
+     "id": "${SCAN_REQUEST_ID}",
```
reasonerjt: See my comments in section `Scanner Adapter API`, IMO, relying on the caller of the API to set request ID is a little problematic

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+     "registry_url": "harbor-harbor-registry:5000",
+     "registry_authorization": "Bearer: JWTTOKENGOESHERE",
+     "artifact_repository": "library/mongo",
+     "artifact_digest": "sha256:917f5b7f4bef1b35ee90f03033f33a81002511c1e0767fd44276d4bd9cd2fa8e"
```
reasonerjt: shall we add a field `Headers:[{$key:$value}, ...]` and remove the `registry_authorization `.  This would be more extensible, considering we may need extra header for Harbor to bypass policies that may block pulling.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+   }
+   EOF
+
+   Status: 202 Accepted
+   ```
+4. Try getting scan report (in unified Harbor format understandable by Harbor Web console):
+   ```
+   curl -H 'Accept: application/vnd.scanner.adapter.vuln.report.harbor.v1+json' \
```
reasonerjt: This is not consistent with the `scanner-adapter-openapi-v1.0.0.yaml`
I prefer the definition in the yaml file, and see the comments in that yaml file for other comment.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+       },
+     ]
+   }
+   ```
+7. Alternatively we could request an original (or raw) vulnerability scan report:
+   ```
+   curl -H 'Accept: application/vnd.scanner.adapter.vuln.report.raw' \
+      http://scanner-adapter:8080/api/v1/scan/${SCAN_REQUEST_ID}/report
+
+   Content-Type: application/vnd.scanner.adapter.vuln.report.raw
```
reasonerjt: The header should be `Accept`?

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
```diff
@@ -0,0 +1,338 @@
...
+          description: A name of the Harbor repository containing the artifact.
+          example: library/oracle/nosql
+        artifact_digest:
+          type: string
+          description: The artifact's digest, consisting of an algorithm and hex portion.
+          example: "sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b"
+    ScanResult:
+      required:
+        - status
+      type: object
```
reasonerjt: How'bout adding field like `artifact` in the `ScanResult` or `HarborVulnerabilityReport` to reflect which artifact was scanned?

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
```diff
@@ -0,0 +1,338 @@
...
+          $ref: '#/components/schemas/HarborVulnerabilityReport'
+        error:
+          $ref: '#/components/schemas/Error'
+      description: |
+        Represents scan result. Since the scanning is done asynchronously, the response
+        might be in the `pending` state. In case of failure the status equals `error` and the
+        corresponding `error` property contains the detailed message describing the problem.
+        If everything is fine and the status equals `done` the `harbor_vulnerability_report` is populated.
+    HarborVulnerabilityReport:
+      type: object
```
reasonerjt: The structure of the `HarborVulnerabilityReport` may result a gap if we want to keep the UI design of Harbor right now.  Right now the `scan_overview` is a colored bar which take the number of components, including the ones with no vulnerabilities(represented with the color of green).  

Is it because some scanner may not have such information in scan result?  I think it's OK, but if so we need a different UI design.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+```go
+package scanner
+
+// RegistrationSettings consists of EndpointURL and optional Authorization properties to test connection
+// with the corresponding Scanner Adapter.
+type RegistrationSettings struct {
+    // A base URL of the Scanner Adapter.
+    EndpointURL    string
+    // An optional value of the HTTP Authorization header sent with each request to the Scanner Adapter API.
+    Authorization  string
```
reasonerjt: At least we need on other option `SkipCertVerify`.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+    // Returns the scanner registration identifier.
+    Create(registration *Registration) (int64, error)
+    // Get returns the details of the specified scanner registration.
+    Get(registrationID int64) (*Registration, error)
+    // Update updates the specified scanner registration.
+    Update(registration *Registration) error
+    // Delete deletes the specified scanner registration.
+    Delete(registrationID int64) (*Registration, error)
+    // SetAsDefault marks the specified scanner registration as default.
+    SetAsDefault(registrationID int64) error
```
reasonerjt: I assume this will unset other registration as `non-default`?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+    // GetDefault returns the default scanner registration or `nil` if there are no registrations configured.
+    GetDefault() (*Registration, error)
+    // Ping pings Scanner Adapter to test EndpointURL and Authorization settings.
+    // Returns `nil` if connection succeeded, a non `nil` error otherwise.
+    Ping(settings *RegistrationSettings) error
```
reasonerjt: What API of Adapter should `Ping` calls?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+3. **Scanner Registry** - An internal component logically responsible for managing the configurations for invoking a
+   scanner adapter
+   1. Backed by persistence in the DB
+   2. Basic CRUD semantics via new Harbor APIs
+4. **Scanner Registration** - A named configuration for invoking a scanner via its adapter
+   1. Name - The name of the entry, must be globally unique in Harbor
+   2. Description - A description of the scanner for human consumption
+   3. Endpoint - The hostname and port to invoke the Adapter API calls against
+   4. Authorization Configuration - The optional value to set in the HTTP `Authorization` header
+   5. NOTE: no plan for preventing duplicate entries based on endpoint etc as such constraints can be delegated to the
```
steven-zou: The endpoint SHOULD be unique and this constraint can be guaranteed by the system/api

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+}
+```
+
+The implementation of the `scanner.Registry` interface will persist `Registration` entities in the Harbor DB.
+There will be a new `scanner_registration` table and the corresponding
+`scanner_registration_id_seq` sequence.
+
+```sql
+create table scanner_registration
+(
```
reasonerjt: Shall we add "type" to help user know what was the vendor of the scanner?
This will potentially be helpful if we want to handle  heterogeneity of the response from different scanner?

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: If the adapter can provide metadata info API, maybe some necessary info can be retrieved from that API and then avoid set them via configuration.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+    Vendor  string
+    Version string
+
+    Capabilities []Capability
+}
+
+// Capability consists of the set of recognized artifact MIME types and the set of scanner report MIME types.
+// For example, a scanner capable of analyzing Docker images and producing a vulnerabilities report recognizable
+// by Harbor web console might be represented with the following capability:
+// - artifact MIME types:
```
reasonerjt: Is there a reference so the MIME type of artifacts can be consistent across scanners?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+   2. Description - A description of the scanner for human consumption
+   3. Endpoint - The hostname and port to invoke the Adapter API calls against
+   4. Authorization Configuration - The optional value to set in the HTTP `Authorization` header
+   5. NOTE: no plan for preventing duplicate entries based on endpoint etc as such constraints can be delegated to the
+      system administrator’s discretion based on deployment specific requirements.
+5. **Scanner Adapter Client** - An HTTP client that takes a registration as configuration input and invokes Scanner
+   Adapter API calls to initiate scans, retrieve results, query health, and refresh metadata from Scanner Adapter(s).
+7. **ScannerJob** - The unit-of-work for executing and retrieving results for a scan of an artifact in the Harbor
+   registry. Returns a ScanResult
+   1. This may be extended to include an aggregation meta-job that defines and executes multiple ScanJobs themselves
```
steven-zou: For executing multiple jobs, that should be put into the `controller` part.

Scan job has a single responsibility, launch scan and waits for the report is ready (by loop) and then notify the interested parties like `controller` with related data via job webhook.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+### Scanner Adapter Client
+
+Given a Scanner Registration settings, it provides functions for accessing the Scanner Adapter API to the ScannerJob.
+Transparently polls the API for getting the results back until the results are ready or an error occurs.
+
+```go
+package scanner
+
+// Metadata represents scanner's metadata and capabilities.
+type Metadata struct {
```
reasonerjt: Currently on UI Harbor will show the timestamp when the Clair vulnerability data was updated from the internet, can we include such information in metadata?  This status is important to the admin IMO.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+      4. Development independent of Harbor processes, no burden on core maintainers
+   2. Disadvantages
+      1. Operational overhead for users to run more services as part of Harbor deployment
+      2. Restricted API has limited data context from Harbor
+      3. API management (versioning etc) required with no ability to update adapters to meet new API changes
+      4. Requires more security considerations because it presents another external attack surface
+2. Harbor defined data model for scan results
+   1. Advantages:
+      1. Common schema needed to support UI visualization
+      2. Can be augmented with scanner-specific data later that could be presented raw to users in UI to avoid
```
reasonerjt: @xaleeks could you confirm we should avoid per-scanner UI work to focus on showing common information?

That means we can only show the minimal set of information across scanners.

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: I think the follow will be:
`harbor API` -> `scan controller` -> `scanner job` -> `scanner client` -> `scanner adapter (service)` -> `scanner`.

core => job service => adapter service => adapter

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: You're right, image layers are pulled by the scanners themselves, harbor only needs to make token for their requests.

proposals/pluggable-image-vulnerability-scanning_proposal.md
reasonerjt: @steven-zou 
I believe you mean 
core => job service => adapter service => **scanner**
correct?

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: If I'm not mistaken, only persist the last copy of the scan report, not all of them (at least in Phase 1).

proposals/pluggable-image-vulnerability-scanning_proposal.md
zhill: I think we should be careful here with complexity in the scanners fetch requirements. Anchore, for example uses fairly standard registry clients so heavy customization of the data fetch requests will be problematic (we use skopeo, specifically since it supports oauth, etc natively) I think the validation point should be will the credentials work for a `docker pull <registry_url>/<project>@<digest>` assuming something like a `docker login` is supported by the scanner.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,707 @@
...
+
+### Work Items
+
+1. Adapter selection and configuration framework
+2. Adapter client implementation
+3. Anchore Engine Adapter
+4. Aqua MicroScanner Adapter
+5. Clair Adapter (?)
+
+## Open issues (if applicable)
```
renmaosheng: Is there a workflow that describes how admin can disable the scanner? I asked because in past experience, Clair has a bug < https://github.com/coreos/clair/issues/593 >, it has a dead loop to retrieve vulnerability data from lots of security website, it leads to the security websites feel harbor is maliciously causing DOS attack. we should at least provide a way to disable the whole scanner if such bad thing happens.

danielpacak: I think that's a very fair point @renmaosheng What's currently described in the proposal does not allow enabling/disabling a Scanner Registration. You could only delete a misbehaving scanner if you will.

I think what we can do is to add an additional property to a Scanner Registration which will indicate whether the scanner is enabled or not. Then the admin could change the value of the flag in the Harbor web console. On the other hand the Scanner Controller will check the flag before scheduling a ScannerJob.

Do you have any other ideas or concerns guys with the enabled/disabled flag? If not I'm happy to update the Scanner Registration entity with such flag.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I came up with updated diagram. Please let me know if it's okey now. If yes I'll proceed with updating description in markdown.

harbor_scanner_adapter_sequence

danielpacak: Sorry for confusion. I've updated the diagrams by including the HLD that used during our scanning group meetings. The sequence diagram is also updated to include Scan Controller and Scan Job.

One more significant change is related to what we discussed in some other thread, i.e. saving the latest scan report in Harbor DB. In short, Scan Job will publish ScanCompleted event. Then there'll be a corresponding handler which in turn persists the ScanReport in the Harbor DB thru Scan Store.

Please review the diagrams one more time. It they're fine please Resolve the conversation @reasonerjt @steven-zou @lizrice

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: You're right @reasonerjt @steven-zou I oversimplified things in the first run. As mentioned in the previous conversation the deployment diagram is now update to the one we've been during our calls.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: As @lizrice mentioned, indeed I wanted to be back-compatible with the current Clair approach where only the DetailsKey (= Clair layer's name) is saved in the Harbor DB. Then the DetailsKey is used to fetch a Scan Report from Clair API.

However, as we agreed, the diagrams are updated to reflect that we're going to save the latest Scan Report for a given Artifact in Harbor DB.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Thanks. Fixed.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Overall I'm okey with Scanner Adapter generating UUID in Phase 1. However my rationale behind such proposal is that HTTP polling might at some point generate unbearable traffic and do not scale well.

To be able to switch painlessly from HTTP to Messaging (AMQP or other pub/sub middleware) I though about [Correlation Identifier EIP](https://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html) and requestor generating request UUID.

In general there are six parts to correlation identifier:
* Requestor - An application that performs a business task by sending a request and waiting for a reply.
* Replier - Another application that receives the request, fulfills it, and then sends the reply. It gets the request ID
from the request ID from the request and stores it as the correlation ID in the reply.
* Request - A Message sent from the requestor to the replier, containing a request ID
* Reply - A Message sent from the replier to the requestor, containing a correlation ID
* Request ID - A token in the request that uniquely identifies the request
* Correlation ID - A token in the reply that has the same value as the request ID in the request

So in our case:

* Requestor is Harbor / Scan Job
* Replier is Scanner Adapter
* Request is a ScanRequest
* Reply is a ScanReport
* Request ID is a UUID v4 generated by Harbor / Scan Job - not an id from DB sequence

Having said that let's simply vote! WDYT? @reasonerjt @steven-zou @lizrice @zhill :

- 👍ScanRequest UUID should be generated by Harbor
- 👎ScanRequest UUID should be generated by Scanner Adapter (in response to ScanRequest and sent in Location HTTP response header as was originally proposed)

(Based on the opinion poll I'll update the proposal respectively.)

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Got your point @reasonerjt Also replied to your first comment regarding ScanRequest ID.
I'm closing this conversation and let's continue discussion there.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Yup. As we agreed we're going to persist the latest Scan Report for a given Artifact. We won't maintain a history of scan reports in phase 1.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Good catch. The OpenAPI spec has been updated.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I agree with @zhill . During the last meeting I rushed into replying yes for additional headers passed to Docker Registry hosted by Harbor. On the other hand it might be problematic for MicroScanner as well. In the POC I used Docker Engine and Docker Client which AFAIK do not allow for passing such custom headers to the registry.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I've updated the example in *.md

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
danielpacak: Added both scanner and artifact object to the definition of HarborVulnerabilityReport. Please double check if I missed something. Otherwise let's resolve the conversation.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
danielpacak: Another good catch @reasonerjt I've brought back ComponentsOverview property/object in HarborVulnerabilityReport which should allow us to render the widget showing the number of vulnerabilities with different severities in the UI.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Agreed. Added

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Yup. Added this assumption to the Godoc / comment.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Yup. Added comment mentioning that the implementation will send a GET request to the /health endpoint.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I've added supported mime types (artifacts / reports) to Scanner Adapter API spec *.yaml.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Good point. Added VulnerabilityDatabase to the ScannerAdapterMetadata object here and OpenAPI spec:

```js
"vulnerability_database": {
  "updated_at": "2019-08-13T08:16:33.345Z"
}
```

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,771 @@
...
+5. The Scan Job instantiates the generic Scanner Adapter Client passing Registry Settings as constructor args.
+6. The Scan Job creates a ScanRequest. (Scan Job is responsible for generating a UUID for the ScanRequest. It's used
+   later on to retrieve the corresponding ScanReport.)
+7. The Scan Job sends the ScanRequest to a Scanner Adapter, which in turn, forwards the request to the underlying Scanner.
+8. At this stage the Scanner pulls the Artifact to perform the actual scan for vulnerabilities and generate the report.
+9. The Scan Job blocks and periodically pulls the ScanReport from the Scanner Adapter until the report is available.
+   Note the identifier of the ScanRequest and the MIME type passed to retrieve a ScanReport.
+10. The Scan Job emits a `ScanCompleted` event.
+11. The `ScanCompleted` event is dispatched to the Event Handler.
+12. The Event Handler saves the ScanReport along with its MIME type and the Artifact's digest to the Harbor DB.
```
steven-zou: As we discussed, Harbor will store the last copy of scanning. For the report, there might be different MIME formats. Shall we retrieve and keep them all or only keep the `harbor format`.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,771 @@
...
+    // A base URL of the Scanner Adapter.
+    EndpointURL    string
+    // An optional value of the HTTP Authorization header sent with each request to the Scanner Adapter API.
+    Authorization  string
+    // SkipCertVerify a flag indicating whether the client should verify the Scanner Adapter's certificate.
+    SkipCertVerify bool
+}
+
+// Registration represents a named configuration for invoking a scanner via its adapter.
+type Registration  struct {
```
steven-zou: I think `Registration` can include `RegistrationSettings` with a flat view as there are not too many properties.

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: Maybe `/v1/metadata` api itself can be used to do`ping`.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,771 @@
...
+   {
+     "scanner": {
+       "name": "Microscanner",
+       "vendor": "Aqua Security",
+       "version": "3.0.5",
+     },
+     "vulnerability_database": {
+       "updated_at": "2019-08-13T08:16:33.345Z"
+     },
+     "capabilities": [
```
steven-zou: Seems the capabilities only include the supported MIME types and result format. What about
- what scan functions supported
  - VUL
  - License
  - Package 
  - etc..
- Whether support policy check or not


If we plan to support those things someday later, we need to consider to provide an extensible data model for easily adding new capability in.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
danielpacak: Harbor won't provide the number of scanned packages/components. We'll just show vulnerable ones.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: We'll use /metadata endpoint . We don't need to support /health

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Check how we configure replication registries and do it consistently.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: This info does not make sense for license scanners or yet-to-define scanners.

proposals/pluggable-image-vulnerability-scanning_proposal.md
zhill: For this proposal I think discussion should be limited to vulnerabilities and policy checks, which are in scope to deliver (policy check optionally). The others should be later proposals that include the specific schema, a mime-type, and the necessary consumption logic. We can clarify what the extension process would look like as an example in this proposal but should avoid any detail until fully delivering such extensions.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.0.yaml
danielpacak: I've updated the proposal in Compatibility section.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Updated in the proposal.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I proposed generic key -value properties incorporating each capability. We might have a properties schema and recognize some of them, e.g.:

```
"harbor.scanner-adapter/scanner-type": "os-package-vulnerability",
"harbor.scanner-adapter/vulnerability-database-updated-at": "2019-08-13T08:16:33.345Z"
```

proposals/pluggable-image-vulnerability-scanning_proposal.md
reasonerjt: Talked to @renmaosheng offline, this will not be covered in this proposal.

It's OK to resolve it.

proposals/pluggable-image-vulnerability-scanning_proposal.md
zhill: Yes, similar to k8s labels/annotations. This seems the most flexible way to present metadata. The consumer (Harbor) will just have to know to look for specific keys based on the type of the scanner if it intends to have logic based on those attributes.

zhill:
I've been thinking about the scan request UUID and generating in client side vs server (scanner adapter) side. I think it should be generated by the adapter. That allows the adapter to generate an id based on its implementation and whether or not it has state. As an example, the generated ID could be a b64 encode of the source request. That allows it to be stateless but still have the id be opaque to the client (harbor). Since the id serves only as a route to identify where to poll for the scan result, the adapter should generate this token in a way that it knows how to resolve. that scheme may vary from adapter to adapter and thus should be generated on the adapter side.

Any information that the harbor client includes that is expected to be returned on subsequent report queries requires some state management in the adapter, and should be minimized or removed if possible. Adapters may have state but we should avoid choices in the API spec that will require it.

lizrice:
There might be some terminology to clear up here. 
* [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) don't encode meaning 
* but maybe we are really talking about a Request ID that just has to be sufficiently unique and doesn't have to be an actual UUID? 

The concern I would have about the adapter generating the ID is whether Harbor needs it to be truly unique across all scanners and across all time. If they are generated by the adapter, Harbor will need to handle the error cases of badly-written adapters that don't actually generate unique IDs.

zhill:
> There might be some terminology to clear up here.
> 
> * [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) don't encode meaning
> * but maybe we are really talking about a Request ID that just has to be sufficiently unique and doesn't have to be an actual UUID?
> 
> The concern I would have about the adapter generating the ID is whether Harbor needs it to be truly unique across all scanners and across all time. If they are generated by the adapter, Harbor will need to handle the error cases of badly-written adapters that don't actually generate unique IDs.

The ID is expected to only be unique to a single adapter since the endpoint to send the requests to has to be known by Harbor. I agree with the meaning confusion and based on our discussions of how the identifier is used, I don't think it needs to be an actual UUID. It is simply an identifier that Harbor can use to request the results of a scan against a specific adapter. If harbor needs additional mapping of job ids or scan ids it can maintain those in the scan job state.

Passing a token or uuid as part of the request (metadata object in json?) to the adapter is fine, but the adapter should define the id for retrieving the results associated with that scan request. This also aligns better with having the adapter control the lifecycle of the id (how long it is valid). If it is a client/harbor generated ID then that lifecycle should be controlled by Harbor (e.g. explicit delete on the id resource once the scan result is successfully retrieved).

steven-zou:
> > There might be some terminology to clear up here.
> > 
> > * [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) don't encode meaning
> > * but maybe we are really talking about a Request ID that just has to be sufficiently unique and doesn't have to be an actual UUID?
> > 
> > The concern I would have about the adapter generating the ID is whether Harbor needs it to be truly unique across all scanners and across all time. If they are generated by the adapter, Harbor will need to handle the error cases of badly-written adapters that don't actually generate unique IDs.
> 
> The ID is expected to only be unique to a single adapter since the endpoint to send the requests to has to be known by Harbor. I agree with the meaning confusion and based on our discussions of how the identifier is used, I don't think it needs to be an actual UUID. It is simply an identifier that Harbor can use to request the results of a scan against a specific adapter. If harbor needs additional mapping of job ids or scan ids it can maintain those in the scan job state.
> 
> Passing a token or uuid as part of the request (metadata object in json?) to the adapter is fine, but the adapter should define the id for retrieving the results associated with that scan request. This also aligns better with having the adapter control the lifecycle of the id (how long it is valid). If it is a client/harbor generated ID then that lifecycle should be controlled by Harbor (e.g. explicit delete on the id resource once the scan result is successfully retrieved).

If it is better to generate UUID at the adapter side, let's do it. Harbor can use its own ID to track the scan job(s).  What do you think?

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: Let's use the enhanced token to bypass the checks. 

Resolve this comment now.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,888 @@
...
+   // Deny pulling artifacts by the client requests originating from the specified source IP.
+   Deny(source IP) error
+}
+```
+
+With such extension, whenever a Scan Registration is created by a Harbor admin, the corresponding allow access rule
+can be created to allow the underlying scanner downloading artifacts from Docker Registry without being blocked by
+the policy checker proxy. When a Scan Registration is deleted the corresponding access rule will be deleted as well.
+
+#### OAuth 2 Bearer Tokens
```
steven-zou: Choose this one. We'll add additional info into the token and parse that token to bypass the checks.

Shall we update the content here?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,888 @@
...
+           "application/vnd.docker.distribution.manifest.v2+json"
+         ],
+         "produces_mime_types": [
+           "application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0",
+           "application/vnd.scanner.adapter.vuln.report.raw"
+         ]
+       }
+     ],
+     "properties": {
+       "harbor.scanner-adapter/scanner-type": "os-package-vulnerability",
```
steven-zou: We need to do adjustments here as we discussed in our last meeting.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,888 @@
...
+    // A base URL of the Scanner Adapter.
+    EndpointURL    string
+    // An optional value of the HTTP Authorization header sent with each request to the Scanner Adapter API.
+    Authorization  string
+    // SkipCertVerify a flag indicating whether the client should verify the Scanner Adapter's certificate.
+    SkipCertVerify bool
+}
+
+// Registration represents a named configuration for invoking a scanner via its adapter.
+type Registration  struct {
```
steven-zou: I'd like to suggest persisting some of the info got from the metadata into the `Registration`, like the `Scanner` info defined in the below section.

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: Can this comment resolved now?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,888 @@
...
+type Registry struct {
+    // A base URL of the Docker Registry v2 API exposed by Harbor.
+    URL           string
+    // An optional value of the HTTP Authorization header sent with each request to the Docker Registry v2 API.
+    // For example, `Bearer: JWTTOKENGOESHERE`.
+    Authorization string
+}
+
+// Artifact represents an artifact stored in Registry.
+type Artifact struct {
```
steven-zou: We can use the `Artifact` defined for the scan controller in the below section for unifying consideration.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,888 @@
...
+type Controller interface {
+    // Scan scans the given artifact.
+    //
+    //  Arguments:
+    //    artifact *Artifact : object includes the kind and metadata of the scanning artifact.
+    //    options  Options   : options for scanning, such as access token of Harbor etc.
+    //
+    //  Returns:
+    //    error  : non-nil error if any errors occurred
+    Scan(artifact *Artifact, options Options) error
```
steven-zou: For the `options`, how about using the following way?

```go
type Option interface {
   Apply(options * Options) error
}

type AuthOption struct {
    auth string
}

func (ao *AuthOption) Apply(options * Options) error {
    options.RegistryAuthorization = ao.auth
    
    return nil
}

Scan(artifact *Artifact, options ...Option) error
```

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Done.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: As discussed lets go for Scanner Adapter generating the identifiers used to fetch corresponding ScanRepots.

danielpacak:
> > > There might be some terminology to clear up here.
> > > 
> > > * [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) don't encode meaning
> > > * but maybe we are really talking about a Request ID that just has to be sufficiently unique and doesn't have to be an actual UUID?
> > > 
> > > The concern I would have about the adapter generating the ID is whether Harbor needs it to be truly unique across all scanners and across all time. If they are generated by the adapter, Harbor will need to handle the error cases of badly-written adapters that don't actually generate unique IDs.
> > 
> > 
> > The ID is expected to only be unique to a single adapter since the endpoint to send the requests to has to be known by Harbor. I agree with the meaning confusion and based on our discussions of how the identifier is used, I don't think it needs to be an actual UUID. It is simply an identifier that Harbor can use to request the results of a scan against a specific adapter. If harbor needs additional mapping of job ids or scan ids it can maintain those in the scan job state.
> > Passing a token or uuid as part of the request (metadata object in json?) to the adapter is fine, but the adapter should define the id for retrieving the results associated with that scan request. This also aligns better with having the adapter control the lifecycle of the id (how long it is valid). If it is a client/harbor generated ID then that lifecycle should be controlled by Harbor (e.g. explicit delete on the id resource once the scan result is successfully retrieved).
> 
> If it is much better to generate UUID at the adapter side, let's do it. Harbor can use its own ID to track the scan job(s). What do you think?

Overall I'm okey with Scanner Adapter generating identifiers for fetching ScanReports. The proposal has been updated to reflect that.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: That's perfectly fine and more extensible way of passing Options to ScanController. In the proposal we can keep it simple, but in the actual PR and prod code I'd definitely take this approach.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Yup. Done.

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: How about persisting the `name`, 'vendor' and 'version' into the database and find a way to refresh them when necessary.

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,875 @@
...
+   4. Authorization Configuration - The optional value to set in the HTTP `Authorization` header
+   5. NOTE: no plan for preventing duplicate entries based on endpoint etc as such constraints can be delegated to the
+      system administrator’s discretion based on deployment specific requirements.
+5. **Scanner Adapter Client** - An HTTP client that takes a registration as configuration input and invokes Scanner
+   Adapter API calls to initiate scans, retrieve results, query health, and refresh metadata from Scanner Adapter(s).
+7. **Scan Job** - The unit-of-work for executing and retrieving results for a scan of an artifact in the Harbor
+   registry.
+   1. This may be extended to include an aggregation meta-job that defines and executes multiple Scan Jobs themselves
+      and combines their results in a specific way
+   2. Identified by a UUID, each job is unique
```
lizrice: I think we agreed that these don't have to be [standard UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) and it's up to the adapter to come up with something that will be unique. To avoid confusion, I think it would be better to say `Identified by a unique ID` rather than UUID here, and describe that the adapter could use a UUID, or some other ID that it guarantees to be unique per job

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,875 @@
...
+![harbor-scanner-adapter-deployment][harbor-scanner-adapter-deployment-img]
+
+![harbor-scanner-adapter-sequence][harbor-scanner-adapter-sequence-img]
+
+1. A User requests a scan of selected Artifact by clicking the Scan button.
+2. The Harbor API forwards the request to the Scan Controller.
+3. The Scan Controller retrieves the default Scanner Registration from the Scanner Registry.
+4. The Scan Controller enqueues a Scan Job with params.
+5. The Scan Job instantiates the generic Scanner Adapter Client passing Registry Settings as constructor args.
+6. The Scan Job creates a ScanRequest. (Scan Job is responsible for generating a UUID for the ScanRequest. It's used
```
lizrice: Isn't this unique ID generated by the adapter now?

proposals/pluggable-image-vulnerability-scanning_proposal.md
```diff
@@ -0,0 +1,875 @@
...
+
+### Scanner Adapter API
+
+Scanner vendors are supposed to implement Scanner Adapters exposing the Scanner Adapter API as specified in
+[Scanner Adapter v1.0 - OpenAPI Specification][scanner-adapter-api-spec].
+
+> Note: OpenAPI spec yaml file can be opened in the online [Swagger Editor][swagger-editor].
+
+- The deployment method is up to the vendor as long as the mounted API endpoint URL is accessible to Harbor services.
+- Scan Job UUID should be a scan request ID used in conjunction with the artifact ID, i.e. artifact repository and
```
lizrice: To avoid confusion I suggest `Scan job ID` rather than UUID throughout this section

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: Agree

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: Generated by the adapter now. Here may need some udpates

proposals/pluggable-image-vulnerability-scanning_proposal.md
steven-zou: +1

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
```diff
@@ -0,0 +1,388 @@
+openapi: 3.0.0
```
steven-zou: no prefix `/api/v1` now?

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
```diff
@@ -0,0 +1,388 @@
...
+      parameters:
+        - name: scan_request_id
+          in: path
+          description: The identifier of the corresponding ScanRequest
+          required: true
+          style: simple
+          explode: false
+          schema:
+            type: string
+            format: uuid
```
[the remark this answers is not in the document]
danielpacak: Good catch. I've removed the format constraint and explained "uniqueness" expectations on identifiers return by AcceptScanRequest operation.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
danielpacak: It's there. Under `.servers/url`.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: I think we agreed to stick to the Authorization header because of existing OCI / docker CLI tooling. Bypassing policy check will be handled with a special access token generated by Harbor.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: We approved the schema for metadata and those generic key-value properties during the last wg meeting.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Here we referred to a Scan Job enqueued internally by Harbor, not a background scanning task enqueued by a Scanner Adapter after accepting a Scan Request. Neither has to be identified by a standard UUID.

I also clarified that in the Scanner Adapter API spec.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: You're right. I've update the proposal to reflect that Scanner Adapters are supposed to generate unique identifiers.

proposals/pluggable-image-vulnerability-scanning_proposal.md
danielpacak: Agree. Clarified.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
```diff
@@ -0,0 +1,386 @@
...
+      example: "3fa85f64-5717-4562-b3fc-2c963f66afa6"
+    Registry:
+      type: object
+      properties:
+        url:
+          type: string
+          description: A base URL or the Docker Registry v2 API exposed by Harbor.
+          format: url
+          example: https://harbor-harbor-registry:5000
+        authorization:
```
steven-zou: This is not correct now. @danielpacak, we need to update it to let the developer know it's a basic auth, not bearer token.

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
```diff
@@ -0,0 +1,392 @@
...
+            "application/vnd.scanner.adapter.error+json; version=1.0":
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /scan:
+    post:
+      tags:
+        - Scanner
+      summary: Accept artifact scanning request
+      description: |
+        A non-blocking operation which enqueues a scan job and returns immediately. It returns a unique
```
zhill: I think we can strike "unique" here to ensure no confusion with uuid. Just say "Returns and Identifier generated by the adapter that can be used to poll for the generated scan reports"

proposals/assets/pluggable-image-vulnerability-scanning/scanner-adapter-openapi-v1.0.yaml
```diff
@@ -0,0 +1,392 @@
...
+          $ref: '#/components/schemas/Artifact'
+    ScanResponse:
+      required:
+        - id
+      properties:
+        id:
+          $ref: '#/components/schemas/ScanRequestId'
+    ScanRequestId:
+      desciption: |
+        A unique identifier returned by the [/scan](#/operation/AcceptScanRequest] operations. The format of the
```
zhill: Same comment here re: remove "unique" to avoid confusion with uuids.

steven-zou:
Hello, @goharbor/all-maintainers @lizrice @michmike  @danfengliu @zhill

As we have got enough votes (4/5) now, this proposal is accepted and merged. Thank you all for drafting, reviewing and voting!

PR #110, opened 2019-10-17
Create CODE_OF_CONDUCT.md

michmike:
@goharbor/all-maintainers please vote on the code of conduct for Project Harbor


renmaosheng:
we also have: https://github.com/goharbor/harbor/blob/master/code-of-conduct.md , may be we can point the same.

michmike:
> we also have: https://github.com/goharbor/harbor/blob/master/code-of-conduct.md , may be we can point the same.

deleted it now

PR #111, opened 2019-10-17
Cloud native registry support proposal

ywk253100:
Signed-off-by: Wenkai Yin <yinw@vmware.com>


proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,375 @@
...
+Authors: 
+* WenkaiYin
+* Alex Xu
+* Steven Ren
+* Daniel Jiang
+* Steven Zou
+* Ziming Zhang
+
+## Abstract
+Proposal is for Harbor to extend support for new cloud-native artifact types such as Helm and CNAB and OPAs and others through following a common set of industry favored APIs (such as OCI standard). All OCI-compliant artifacts should subsequently be able to be onboarded to Harbor if desired. This means that Harbor will host new artifact types and correctly support all corresponding actions of all various artifacts based on `artifact type` such as when they need to be pushed, discovered through catalog & search APIs, scanned, pulled, replicated, webhooks being fired as a results etc. This is largely an engineering refactor/remodularization effort that allows Harbor to stay ahead and quickly absorb new cloud-native artifact types as they gain traction in the market. Support for these artifacts can be just as easily removed from Harbor as it is added.
```
wy65701436: please gives some background for OCI.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+
+![overall_architecture](images/cloud-native-registry-support/overall_architecture.png)
+
+### API
+
+#### Pattern
+
+| Function | v1(Current) | v2 | Request | Response| Comment |
+| -------- | ------------| -- | ------- | ------- | ------- |
+| List artifacts | GET /api/repositories/{repository_name}/tags | GET /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts | Query string:  label=true/false(default): return label or not  signature=true/false(default): return signature or not  vulnerability=true/false(default): return vulnerability or not  type=image/chart  tag=xxx  label_id=1  page=1  size=10 | Artifact list |  |
```
steven-zou: Use the more meaningful name to replace the id?

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+```go
+type Artifact struct {
+	ID                   int64                  `json:"id"`
+	Type                 string                 `json:"type"`       // index, image or chart
+	MediaType            string                 `json:"media_type"` // the specific media type
+	Repository           *Repository            `json:"repository"`
+	Digest               string                 `json:"digest"`
+	Tags                 []*Tag                 `json:"tags"` // the list of tags that attached to the artifact
+	Size                 int64                  `json:"size"`
+	Uploader             *User                  `json:"uploader"` // the user who uploads this artifact, this is needed when the artifact is an untagged one
```
steven-zou: `Uploader` should not be in the `Artifact` model

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+	UploadTime           time.Time              `json:"upload_time"`
+	VulnerabilitySummary *VulnerabilitySummary  `json:"vulnerability_summary"` // only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
+	Labels               []*Label               `json:"labels"`                // this is designed to keep some of the big fragment info which is also extracted from the artifact content and stable: build history for image, values, readme and dependencies for chart
+	ExtraAttrs           map[string]interface{} `json:"extra_attrs"`
```
steven-zou: `ExtraAttrs` keeps k-v property pairs extracted from the config file of the artifact, which may be different among different artifacts.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+	Additions            []string               `json:"additions"`
```
steven-zou: `Additions` contains the info from other configurable services, like `Scan Report` or `Signature`.

proposals/cloud-native-registry-support.md
heww: Suggest to use fields to limit which fields are returned by the API

https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#limiting-fields

https://open.taobao.com/api.htm?docId=21349&docType=2

proposals/cloud-native-registry-support.md
ywk253100: I'm considering why we don't merge this two properties `ExtraAttrs` and `Additions`? The types of them are both `map[string]interface{}`. Some of attributes are got from database(such as os, architecture) and some are got from other service(such as scan report and signature). We can put all of them into a single `ExtraAttrs` when assembling it. This can reduce the confusion. And the consumer of `Artifact` doesn't care there is only one `ExtraAttrs` or two properties.

proposals/cloud-native-registry-support.md
ywk253100: @heww Some of the attributes are not just simple fields, such as signature and vulnerability, them are contained in the `Additions` field. And we're not suppose to support specifying every field.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+| Get an artifact | GET /api/repositories/{repository_name}/tags/{tag} | GET /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id} | Query string:  label=true/false(default)  signature=true/false(default)  vulnerability=true/false(default) | Artifact |  |
+| Delete an artifact | DELETE /api/repositories/{repository_name}/tags{tag} | DELETE /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id} |  |  |  |
+| Delete a tag | None | DELETE /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/{tag} |  |  | This API provides the capability that deleting one tag will not delete other tags even though the digests of them are same |
+| Retag/Copy from | POST /api/repositories/{repository_name}/tags | POST /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts | Query string: from=source_artifact_id  Request Body:  { "tags": ["1.0", "latest"]} |  | When the query string`from` presents, this endpoint is used to copy, else it is used to upload(not support currently). Only works for image. The chart name and version is recorded in "Chart.yaml" |
```
bitsf: /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/copy

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+    "detail": "the name of project must contains 2 characters at least, and ..."
+}
+```
+For one kind of error we should use one same error code, message and different detail.  
+Such as for HTTP 400 error, may contain the following specified error:  
+* invalid json
+* invalid field
+* field missing
+
+For 415:
```
bitsf: error code should be string, which is more flexible.
such as
100000415 means harbor defined error, quota exceeded
200000100 means database error

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+* field missing
+
+For 415:  
+* quota exceeded
+* content trust policy violated
+* vulnerability policy violated
+
+#### Versioning
+The APIs are grouped by resource, and each group maintains their own versions and just maintains one version in the same time. Such as:  
+Artifact API group: /api/v2.0/artifacts
```
bitsf: specify which characters are allowed in version.
is "v2.0-beta" allowed

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+	ExtraAttrs           map[string]interface{} `json:"extra_attrs"`// only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
+	Additions            []string     // this is designed to keep some of the big fragment info which is also extracted from the artifact content and stable: build history for image, values, readme and dependencies for chart          `json:"additions"`
+	References           []int64                `json:"references"` // contains the ID list referenced by the artifact if the artifact is an index
+}
+
+// Tag belongs to one repository and can only be attached to a single one artifact under the repository
+type Tag struct {
+	ID                 int64       `json:"id"`
+	Name               string      `json:"name"`
+	Repository         *Repository `json:"repository"`
```
bitsf: This Repository is duplicate with parent, and it would make the json quite big

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+	MediaType            string                 `json:"media_type"` // the specific media type
+	Repository           *Repository            `json:"repository"`
+	Digest               string                 `json:"digest"`
+	Tags                 []*Tag                 `json:"tags"` // the list of tags that attached to the artifact
+	Size                 int64                  `json:"size"`
+	Uploader             *User                  `json:"uploader"` // the user who uploads this artifact, this is needed when the artifact is an untagged one
+	UploadTime           time.Time              `json:"upload_time"`
+	Labels               []*Label               `json:"labels"`                
+	ExtraAttrs           map[string]interface{} `json:"extra_attrs"`// only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
+	Additions            []string     // this is designed to keep some of the big fragment info which is also extracted from the artifact content and stable: build history for image, values, readme and dependencies for chart          `json:"additions"`
```
bitsf: Additions should also be a map[string]interface{} , or it will be difficult to get the keys

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+    * dependencies
+    * readme
+
+#### Data Level model
+```golang
+type Artifact struct {
+	ID           int64     `orm:"pk;auto;column(id)"`
+	Type         string    `orm:"column(type)"`       // index, image or chart
+	MediaType    string    `orm:"column(media_type)"` // the specific media type
+	RepositoryID int64     `orm:"column(repository_id)"`
```
bitsf: is RepositoryID the ProjectID ?

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+}
+
+type Tag struct {
+	ID                 int64     `orm:"pk;auto;column(id)"`
+	RepositoryID       int64     `orm:"column(repository_id)"` // tags are the resources of repository, one repository only contains one same name tag
+	ArtifactID         int64     `orm:"column(artifact_id)"`   // the artifact ID that the tag attaches to, it changes when pushing a same name but different digest artifact
+	Name               string    `orm:"column(name)"`
+	UploaderID         int64     `orm:"column(uploader_id)"` // the user who uploads the tag
+	UploadTime         time.Time `orm:"column(upload_time)"`
+	LatestDownloadTime time.Time `orm:"column(latest_download_time)"`
```
bitsf: We may have a download log table to record the detail download info, in order that we can show download count and other statistics info in future

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+}
+
+type Reference struct {
+	ID          int64 `orm:"pk;auto;column(id)"`
+	ArtifactID  int64 `orm:"column(artifact_id)"`
+	ReferenceID int64 `orm:"column(reference_id)"`
```
bitsf: TagID

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+	// List artifacts according to the query
+	List(query *model.Query, option *model.Option) ([]*model.Artifact, error)
+	// Get the artifact specified by the ID
+	Get(id int64, option *model.Option) (*model.Artifact, error)
+	// Create the artifact specified by the digest under the repository,
+	// returns the artifact ID and error. If tags are provided, attach them
+	// to the artifact.
+	Create(repository *model.Repository, digest string, uploader *model.User, tags ...string) (int64, error)
+	// Attach the tags to the artifact. If the tag doesn't exist, will create it.
+	// If the tag is already attached to another artifact, will detach it first
```
bitsf: I think should not use detach but update, in order to make sure the transaction

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,392 @@
...
+	UploadTime   time.Time `orm:"column(upload_time)"`
+	ExtraAttrs   string    `orm:"column(extra_attrs)"`
+}
+
+type Tag struct {
+	ID                 int64     `orm:"pk;auto;column(id)"`
+	RepositoryID       int64     `orm:"column(repository_id)"` // tags are the resources of repository, one repository only contains one same name tag
+	ArtifactID         int64     `orm:"column(artifact_id)"`   // the artifact ID that the tag attaches to, it changes when pushing a same name but different digest artifact
+	Name               string    `orm:"column(name)"`
+	UploaderID         int64     `orm:"column(uploader_id)"` // the user who uploads the tag
```
bitsf: we should introduce a State column in order to record the upload state, or it will be hard to deal with the transaction problem

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,421 @@
...
+}
+```
+
+### Artifact Manager
+Artifact Manager is designed to maintain the metadata of all the cloud-native artifacts no matter what original types they're before as a unique core component. All the cloud-native artifacts will be abstracted to unified logical objects in the harbor side and managed by this manager.  
+
+Interface definitions for the artifact manager:
+```golang
+type Manager interface {
+	List(*model.Query) ([]*model.Artifact, error)
```
bitsf: should return the total count

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,423 @@
...
+	// As the deletion operation can be perfromed via registry API directly,
+	// here we just delete the artifact from database, not from registry
+	Delete(id int64) error
+	// GetAddition returns the addition content of the artifact
+	// The addition is different according to the artifact type: build history for
+	// image; values.yaml, readme and dependencies for chart
+	GetAddition(id int64, addition string) ([]byte, error)
+	// Prune removes the useless artifact records in database. An useless artifact
+	// is the artifact referenced by no tags or other artifacts(index)
+	Prune() error
```
bitsf: how about add a param repository id, in order to support jobservice tasks

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Proposal
+An artifact managed by Harbor can contain multiple tags which references the same digest manifest. Some APIs and operations get updated based on this.  
+
+For clarification, the following items are all artifacts, `goharbor` is the project part, `core` is the repository part, `sha256:426dxxx...`is the digest and the `1`, `1.0`, `1.0.0`, `latest`, `2.0` are the tags:
+* `goharbor/core@sha256:426dxxx...`
+* `goharbor/core:[1, 1.0, 1.0.0, latest, ...]`(the digest for all tags are same) 
+* `goharbor/core:2.0`(the digest is different with `goharbor/core:1`)
+
+The `goharbor/core@sha256:426dxxx...` has no tags, it's `untagged` artifact; the `goharbor/core:[1, 1.0, 1.0.0, latest, ...]` and `goharbor/core:2.0` have one or more tags, they are `tagged` artifacts. Harbor only manages the tagged ones.
```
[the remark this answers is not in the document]
ywk253100: Remove the `Harbor only manages the tagged ones`

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,422 @@
...
+ 
+// Register resolver
+func Register(resolver ResolverFactory) error {}
+ 
+// List all the resolvers
+func List() []ResolverFactory {}
+```
+
+#### Back off
+To avoid the case that the metadata in the database is out of sync with the real content in the registry, a back off solution is provided. The overall flow can be described with the following diagram.
```
bitsf: we should also have a force sync all back off, which will first delete all database records

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+| Get addition properties for different artifact type(image: build history, chart: values, readme, dependencies) | None | GET /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/additions/{build_history, values, readme, dependencies} |  |  | Index has no build history |
+| Get all signatures of a repository | GET /api/repositories/{repository_name}/signatures | REMOVE |  |  |  |
+| Get top repositories | GET /api/repositories/top | REMOVE |  |  |  |
+| Get manifest | GET /api/repositories/{repository_name}/tags/{tag}/manifest | REMOVE |  |  |  |
+
+#### Error
+The API error consists of two parts:
+* HTTP Response Code
+* Error Payload 
+    * code: The integer error code
```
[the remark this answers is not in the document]
ywk253100: Done

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,382 @@
...
+* Develop new client for upload/download artifacts (this is on artifact authors).
+* Introduce a new standard to store artifacts.  We are going with OCI specification based on current momentum and are not in the business of building standards but active participate in discussions
+* Introduce new format for packaging / bundling artifacts (such as operator).  
+* Serving artifacts that cannot be stored in OCI registry or other general content (e.g. rpm, maven, npm files), with potential to expand to additional file types if needed
+* Client tooling that interfaces w/ Harbor (for push pull etc)
+
+## Proposal
+An artifact managed by Harbor can contain multiple tags which references the same digest manifest. Some APIs and operations get updated based on this.  
+
+For clarification, the following items are all artifacts, `goharbor` is the project part, `core` is the repository part, `sha256:426dxxx...`is the digest and the `1`, `1.0`, `1.0.0`, `latest`, `2.0` are the tags:
```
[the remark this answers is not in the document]
ywk253100: Post a multiple section sample

proposals/cloud-native-registry-support.md
ywk253100: It's repository ID, one artifact belongs to one repository and one repository belongs to one project

proposals/cloud-native-registry-support.md
ywk253100: You mean the access log? I think it already exists

proposals/cloud-native-registry-support.md
ywk253100: It references an artifact, not a tag

proposals/cloud-native-registry-support.md
ywk253100: You mean something like revision?

proposals/cloud-native-registry-support.md
ywk253100: The "Back off" is outdate, I'll remove it. Please refer to "Take Database Data As Final" section.

proposals/cloud-native-registry-support.md
ywk253100: A query string `from` is introduced in this API, when it appears this endpoint is used to copy. This is a similar way with docker registry mount blob API. What do you think about the new way?

proposals/cloud-native-registry-support.md
ywk253100: The 415 here is the HTTP status code

proposals/cloud-native-registry-support.md
ywk253100: Didn't get your point

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+
+![get artifact workflow](images/cloud-native-registry-support/list-get-artifact-workflow.png)
+
+Step 1: the `Server` receives the request  
+Step 2: the `Artifact Controller` assembles the artifact information from different modules: metadata from `Artifact Manager`(step 3,4,5), scanning summary/detail from `Scan` module, sign information from `Sign` module. The scanning and sign information are optional according to the request   
+
+#### Garbage Collection(Harbor API)
+Three action items need to be done:
+* Clean up the useless artifact records in database. An useless artifact is the artifact referenced by no tags or other artifacts(index). It appears when pushing a same name but different digest artifact
+* Remove the artifacts that exist in docker registry but not exist in database from docker registry. This kind of artifacts appear when calling Harbor API to delete tag
```
reasonerjt: More details needed, for example, we only remove manifest in this step, let's explicitly point out which api of the registry will be called. 
So we only need to walk through DB and call API to remove the  manifests that have no tags.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+| API | Function | Comment |
+| ------------ | ------| ------ |
+| **Re-implement** | | |
+| GET /v2/_catalog | List repositores | Return the repositories based on database |
+| GET /v2/{name}/tags/list | List tags | Return the tags based on database |
+| DELETE /v2/{name}/manifests/{reference} | Delete the manifest | Delete the artifact record in the database direclty |
+| **Intercept** | | |
+| GET /v2/{name}/manifests/{reference} | Pull the manifest | Check the existence of artifact before proxying |
+| HEAD /v2/{name}/manifests/{reference} | Check the existence of the manifest | Check the existence of artifact before proxying |
+| **Proxy directly** | | |
```
reasonerjt: Seems the `Proxy directly` and `Intercept` can be merged, actually for the requests under `Proxy directly` are intercepted anyway.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+    * dependencies
+    * readme
+
+#### Data Level model
+```golang
+type Artifact struct {
+	ID           int64     `orm:"pk;auto;column(id)"`
+	Type         string    `orm:"column(type)"`       // index, image or chart
+	MediaType    string    `orm:"column(media_type)"` // the specific media type
+	RepositoryID int64     `orm:"column(repository_id)"`
```
heww: Suggest to save project id in this table.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+	ID                 int64     `orm:"pk;auto;column(id)"`
+	RepositoryID       int64     `orm:"column(repository_id)"` // tags are the resources of repository, one repository only contains one same name tag
+	ArtifactID         int64     `orm:"column(artifact_id)"`   // the artifact ID that the tag attaches to, it changes when pushing a same name but different digest artifact
+	Name               string    `orm:"column(name)"`
+	UploaderID         int64     `orm:"column(uploader_id)"` // the user who uploads the tag
+	UploadTime         time.Time `orm:"column(upload_time)"`
+	LatestDownloadTime time.Time `orm:"column(latest_download_time)"`
+}
+
+type Reference struct {
```
heww: Do we need reference in database? Reference is sha256 digest or tag of the artifact. The sha256 digest was saved in `Artifact` and tag was saved in `Tag`.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+	ID          int64 `orm:"pk;auto;column(id)"`
```
wy65701436: the table name is hard to understand that it's stand for the index refernce.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,413 @@
...
+ 
+// Register resolver
+func Register(resolver ResolverFactory) error {}
+ 
+// List all the resolvers
+func List() []ResolverFactory {}
+```
+
+#### Adapter
+We may need the adapters to translate the V1 APIs to use the new data.
```
wy65701436: can you give more explanation about it?

proposals/cloud-native-registry-support.md
bitsf: it seems not quite clear to just using "from".
mount is actually also the create method by docker push, not mean copy.

proposals/cloud-native-registry-support.md
bitsf: I mean record download log in database

proposals/cloud-native-registry-support.md
bitsf: yes, revision is also ok

proposals/cloud-native-registry-support.md
bitsf: Prune(repository_id)
that we can multi-thread clean up

proposals/cloud-native-registry-support.md
ywk253100: This is what the access log do

proposals/cloud-native-registry-support.md
ywk253100: The `Intercept` here means we need to check the existence of the manifest in database to implement the OCI registry API spec.

proposals/cloud-native-registry-support.md
ywk253100: Why we need the project ID here? It can be get via the repository

proposals/cloud-native-registry-support.md
ywk253100: As we keep data in two locations(database and registry), although we take database data as final, there are cases that the data between database and registry is inconsistent, such as the the artifact is stored in registry but failed to be recorded in database, so we need a way to do the sync. I don't think this can be avoid.

proposals/cloud-native-registry-support.md
ywk253100: I'll rename the struct, it is used for index to record the referenced manifests

proposals/cloud-native-registry-support.md
bitsf: but we can't easily get such info from log, and show them on ui

proposals/cloud-native-registry-support.md
ywk253100: Removed

proposals/cloud-native-registry-support.md
ywk253100: Removed

proposals/cloud-native-registry-support.md
ywk253100: Added

proposals/cloud-native-registry-support.md
ywk253100: Add comment to explain what it is used for

proposals/cloud-native-registry-support.md
ywk253100: Add comment to explain what it is used for

proposals/cloud-native-registry-support.md
ywk253100: As we discussed offline, revision isn't needed.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+## Goals
+* Make Harbor compatible with OCI registry
+* Be able to onboard any other OCI artifact if desired
+* Support the popular registry client:
+  * Docker
+  * Kubelet
+* The tag can be deleted separately without deleting all other sharing-digest tags
+* Replication/quota/tag retention/scan/immutable tag/etc can work as expected after introducing the new artifacts
+* Make sure the performance is in the same level with the current implementation
```
renmaosheng: 'make sure performance and scale doesn't have any regression from v1.10 release'

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+## Background
+Registries started out as a content store for just docker images but are quickly evolving to support other file formats as new cloud native artifacts thrive. Besides being able to support files of varying sizes and formats, registries should have a secure, layered, cross-referenceable file system that can easily expand to support new cloud native artifacts. 
+
+## Goals
+* Make Harbor compatible with OCI registry
+* Be able to onboard any other OCI artifact if desired
+* Support the popular registry client:
+  * Docker
+  * Kubelet
```
renmaosheng: Could we just use Kubernetes? the image pull logic depends on the container runtime plugin used.

proposals/cloud-native-registry-support.md
ywk253100: @steven-zou As we discussed, the name cannot be used when supporting globalization, so we'll use the numeric ID

proposals/cloud-native-registry-support.md
ywk253100: `beta` is not allowed

proposals/cloud-native-registry-support.md
ywk253100: Updated

proposals/cloud-native-registry-support.md
ywk253100: @renmaosheng I updated it to `The container runtimes used by Kubernetes to pull images: https://kubernetes.io/docs/setup/production-environment/container-runtimes/`, is that make sense?

ywk253100:
@renmaosheng The clarification for supporting chart museum is added, see the `Helm chart repository supporting` section for the detail

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+Registries started out as a content store for just docker images but are quickly evolving to support other file formats as new cloud native artifacts thrive. Besides being able to support files of varying sizes and formats, registries should have a secure, layered, cross-referenceable file system that can easily expand to support new cloud native artifacts. 
+
+## Goals
+* Make Harbor compatible with OCI registry
+* Be able to onboard any other OCI artifact if desired
+* Support the popular registry client:
+  * Docker
+  * The container runtimes used by Kubernetes to pull images: https://kubernetes.io/docs/setup/production-environment/container-runtimes/
+* The tag can be deleted separately without deleting all other sharing-digest tags
+* Replication/quota/tag retention/scan/immutable tag/etc can work as expected after introducing the new artifacts
```
xaleeks: replace this with 'all existing features of harbor', there are too many to list

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+* field missing
+
+For 415:  
+* quota exceeded
+* content trust policy violated
+* vulnerability policy violated
+
+#### Evolution
+All APIs will bump up to 2.0 and will break the forward compatibility with the APIs in Harbor 1.x. And we won't maintain the 1.x APIs in Harbor 2.x anymore.  
+If any new break change is introduced to any API in 2.x, bump up the minor version of all APIs to the version of Harbor that the change is made in.
```
xaleeks: please provide an example so everyone know what is meant here

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+* invalid json
+* invalid field
+* field missing
+
+For 415:  
+* quota exceeded
+* content trust policy violated
+* vulnerability policy violated
+
+#### Evolution
```
xaleeks: @stuclem  can you help word this a little better?

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+If any new break change is introduced to any API in 2.x, bump up the minor version of all APIs to the version of Harbor that the change is made in.  
+We'll maintain the latest two versions of APIs with the same major version in the future releases. 
+The version doesn't change if the change breaks nothing.  
+e.g. The API version will be 2.0 when Harbor 2.0 is released. If there is no change during Harbor 2.1, the version of API will still be 2.0. When a break change is introduced to artifact API in Harbor 2.2, all APIs will bump up to version 2.2. So in Harbor 2.2, there will be two version of API: 2.0 and 2.2, and the API 2.0 will be marked as being deprecated and will be removed once new version of API comes out.  
+If we cannot guarantee the support for the latest two versions of API when making the break change, the change should be introduced in a new major version of Harbor and the major version of API will be bumped up as well.  
+
+#### Adapter
+We may need the adapters to translate the V1 APIs to use the new data.
+
+### Take Database Data As Final
```
xaleeks: why does this happen and how is this resolved to be eventually consistent?

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+One big change made in this proposal is that **take database data as final**. That means if there is any inconsistence between database and registry, take database data as final. Such as: 
+* When pushing an image, the image is put successfully in the registry but the artifact record isn't inserted into the database, the operation fails
```
xaleeks: I'm not sure I understand, can you elaborate?

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+* When deleting an artifact, we can just delete the database record, and the real data in the registry will be cleaned up when running garbage collection
```
xaleeks: also not clear what is meant here

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+```go
+type Artifact struct {
+	ID            int64                      `json:"id"`
+	Type          string                     `json:"type"` // image, chart, etc
+	Repository    *Repository                `json:"repository"`
+	Tags          []*Tag                     `json:"tags"`       // the list of tags that attached to the artifact
+	MediaType     string                     `json:"media_type"` // the specific media type
+	Digest        string                     `json:"digest"`
+	Size          int64                      `json:"size"`
+	UploadTime    time.Time                  `json:"upload_time"`
```
xaleeks: I dont think uploadtime is a property of the artifact, maybe creation time but upload time?? what's reasoning here

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+* vulnerability policy violated
+
+#### Evolution
+All APIs will bump up to 2.0 and will break the forward compatibility with the APIs in Harbor 1.x. And we won't maintain the 1.x APIs in Harbor 2.x anymore.  
+If any new break change is introduced to any API in 2.x, bump up the minor version of all APIs to the version of Harbor that the change is made in.  
+We'll maintain the latest two versions of APIs with the same major version in the future releases. 
+The version doesn't change if the change breaks nothing.  
+e.g. The API version will be 2.0 when Harbor 2.0 is released. If there is no change during Harbor 2.1, the version of API will still be 2.0. When a break change is introduced to artifact API in Harbor 2.2, all APIs will bump up to version 2.2. So in Harbor 2.2, there will be two version of API: 2.0 and 2.2, and the API 2.0 will be marked as being deprecated and will be removed once new version of API comes out.  
+If we cannot guarantee the support for the latest two versions of API when making the break change, the change should be introduced in a new major version of Harbor and the major version of API will be bumped up as well.  
+
```
xaleeks: also should make it clear that it's just the version # that changes and not he name, ie basically the change will be from /api/harbor/v1.0/retag to /api/harbor/v2.0/retag  and not to /api/harbor/v2.0/ichangedthenamehere

xaleeks:
@cd1989 @mmpei @kofj  We've advertised this on slack and elsewhere but I haven't seen any comments from you guys yet, so please take time to look through this proposal if you haven't already.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+## Background
+Registries started out as a content store for just docker images but are quickly evolving to support other file formats as new cloud native artifacts thrive. Besides being able to support files of varying sizes and formats, registries should have a secure, layered, cross-referenceable file system that can easily expand to support new cloud native artifacts. 
+
+## Goals
+* Make Harbor compatible with OCI registry
```
xaleeks: should clarity what this means and which specs the Harbor registry need to implement, ie there are several like distribution spec, runtime spec, image spec for ex.  What does registry need to follow and what do artifact authors need to follow.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+We'll maintain the latest two versions of APIs with the same major version in the future releases. 
+The version doesn't change if the change breaks nothing.  
+e.g. The API version will be 2.0 when Harbor 2.0 is released. If there is no change during Harbor 2.1, the version of API will still be 2.0. When a break change is introduced to artifact API in Harbor 2.2, all APIs will bump up to version 2.2. So in Harbor 2.2, there will be two version of API: 2.0 and 2.2, and the API 2.0 will be marked as being deprecated and will be removed once new version of API comes out.  
+If we cannot guarantee the support for the latest two versions of API when making the break change, the change should be introduced in a new major version of Harbor and the major version of API will be bumped up as well.  
+
+#### Adapter
+We may need the adapters to translate the V1 APIs to use the new data.
+
+### Take Database Data As Final
+One big change made in this proposal is that **take database data as final**. That means if there is any inconsistence between database and registry, take database data as final. Such as:
```
cd1989: What's the main purpose of this, some users will copy registry data to registry directly and sync them to database.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+| Delete label from an artifact | DELETE GET /api/repositories/{repository_name}/tags/{tag}/labels/{labels_id} | DELETE /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/labels/{label_id} |  |  |  |
+| Scan an artifact | POST /api/repositories/{repository_name}/tags/{tag}/scan | POST /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/scan |  |  |  |
+| Get vulnerabilities of an artifact | GET /api/repositories/{repository_name}/tags/{tag}/vulnerability/details | GET /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/vulnerabilities |  |  |  |
+| Get addition properties for different artifact type(image: build history, chart: values, readme, dependencies) | None | GET /api/v2/projects/{project_id}/repositories/{repository_id}/artifacts/{id}/additions/{build_history, values, readme, dependencies} |  |  | Index has no build history |
+| Get all signatures of a repository | GET /api/repositories/{repository_name}/signatures | REMOVE |  |  |  |
+| Get top repositories | GET /api/repositories/top | REMOVE |  |  |  |
+| Get manifest | GET /api/repositories/{repository_name}/tags/{tag}/manifest | REMOVE |  |  |  |
+
```
jwangyangls: Cannot  find "/api/project_name/{project_name}/repositories/{repository_name}/artifact/{artifact_id}/scan/{report_id}/log"

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,462 @@
...
+Step 3, 4: clean up the useless artifacts in database 
+Step 5: remove the artifacts that exist in docker registry but not exist in database from docker registry  
+Step 6: several other works during the GC  
+Step 7: clean up the useless blobs in registry  
+
+### Model Definition
+#### Business Logic Level Model
+```go
+type Artifact struct {
+	ID            int64                      `json:"id"`
```
jwangyangls: need to add scanOverview

jwangyangls: the result of scan

renmaosheng:
Let's merge this? the 2.0 is already released, we need to resolve this ASAP.

xaleeks:
agreed, lets merge it? @ywk253100 @reasonerjt

hainingzhang:
LGTM.

proposals/cloud-native-registry-support.md
```diff
@@ -0,0 +1,434 @@
...
+User API group: /api/v3.0/users  
+
+The version consists of two parts: main version and sub version. The version has to be bumped up as long as any changes:
+* Bump up the main version if contains breaking changes
+* Bump up the sub version if contains no breaking changes
+
+The artifact API group will be /api/v2.1/artifacts after bumping up the version including the non-breaking changes.
+
+### Take Database Data As Final
+One big change made in this proposal is that **take database data as final**. That means if there is any inconsistence between database and registry, take database data as final. Such as:
```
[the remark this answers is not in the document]
xaleeks: agreed

steven-zou:
Merge this PR according to the [lazy-consensus](https://github.com/goharbor/community/blob/master/GOVERNANCE.md#lazy-consensus) rule

PR #124, opened 2020-02-26
Enable / Disable Anonymous Access

cafeliker:
Propose a new feature of allowing administrator enable and disable anonymous access through the system settings page.

More discussion is at https://github.com/goharbor/harbor/issues/10760.

Thanks
ye


cafeliker:
@steven-zou can you review this PR?

Thanks
ye

xaleeks:
proposal looks good, please check implementation.  @steven-zou @reasonerjt

reasonerjt:
In regards to the workflow, IMO it's more intuitive if we enhance it at the project level, i.e. when admin sets a project as public, he has additional options to make the project accessible to anonymous users or not.

cafeliker:
@reasonerjt cyber security would ask for the control on the application level, rather than rely on individual's decision.

reasonerjt:
So I see the key disagreement is that should the system admin determine the access policy of a project or a project admin.

Let's keep this PR pending until we reach agreement.

phin1x:
Is der any progress on this pr?

tianon:
I think the system admin should control whether or not it's possible for a project admin to make things available without authentication, but then leave it up to each project to decide whether they want the project "private", "internal", or "public" (which are the words other similar tools like GitLab use to differentiate these three states).

For comparison, GitLab's per-project settings:

image

And the system administrator settings:

image

The reason I think this split makes sense is that the system administrator is typically responsible for determining/enforcing the overall enterprise security policy where the project administrator's responsibilities are going to almost always be a subset of that larger policy.  So IMO, we really need _both_, not just one or the other.

OrlinVasilev:
looks like merged and and closed as PR!

slushysnowman:
This looks like it's been sitting here for a while, but is a pretty important proposal for enterprise.

I'd especially support the suggestion @tianon made - for our requirements as system admins we want to be able to restrict anonymous access entirely, as it doesn't fit our use case - but we do want project admins to be able to set their project as 'internally public' to authenticated users, for purposes of sharing images with other teams within our organization etc.

But I'd also support this proposal as described, as an MVP, as at the moment the current lack of any option to disable anonymous access is pretty scary.

Vad1mo:
I also support @tianon [proposal](https://github.com/goharbor/community/pull/124#issuecomment-824087466) much like GitLab dues it. Private, internal, public.

Vad1mo (approved):
support for  "private", "internal", or "public"

OrlinVasilev:
+1

tillepille:
Our company is also interested in the feature.

With the docker hub rate limiting we've set up a proxy project and want every user (and project robot accounts in best) to be able to pull images through the proxy. 

Can I help somehow?

Vad1mo:
> Can I help somehow?
we need someone who takes the lead in this proposal and its implementation. 

@tillepille, if interested, I can guide you around the request and PR.

tillepille:
@Vad1mo thanks!
I can do my best do lead the proposal but for implementing... I am not really proficient in Go, I fear 😬

Vad1mo:
I am closing this in favor for https://github.com/goharbor/community/pull/242

PR #133, opened 2020-03-25
add proposal of non-blocking GC

wy65701436:


proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Non-Blocking
+
+As a system admin, you configure Harbor to run a garbage collection job on a fixed schedule. At the scheduled time, Harbor:
+
+    Identifies and marks unused image layers.
+    Deletes the marked image layers.
+
+### Mark
+Bases on the Harbor DB, we can count each blob/manifest's reference count, and select the reference count 0 as the candidate.
```
reasonerjt: Before Mark, there is an optional step to remove the untagged artifacts.  Could you also consider that step in the proposal?  
Will there be any race-condition in that step?  
If the answer is no, could you explain in the proposal why not?

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+![mark](../images/non-blocking-gc/mark.png)
+
+
+#### Question 1, how to deal with the uploading blobs at the phase of marking.
+We do have a table to record the uploading blobs info, that's project & blob.
+The delete candidate excludes all of blobs that in the project & blob.
+
+1. candidate set 1 -- all blobs from table blob exclude the items in the table project & blob.
+2. candidate set 1 excludes all of referenced blobs (artifact -> artifact & blob -> blob).
```
reasonerjt: each artifact must belong to a project, so then why there are "referenced blobs" not in "project blobs"?

This is not strictly in the scope of the non-blocking GC but looks like there's redundancy in the schema.

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+
+![mark_uploading](../images/non-blocking-gc/mark_uploading.png)
+
+### Sweep
+The registry controller will grant the capability of deleting blob & manifest.
+
+#### Question 1, how to deal with the uploading blobs at the phase of sweeping.
+Docker client will send a head request to ask the existence of the blob, we will intercept that request.
+
+1. if the blob is in deleting status, fail the request with 404. Docker client will continue to put the blob to registry, but at this time,
```
reasonerjt: What's the difference between deleting and delete status and how do you check the status?

proposals/new/Non-Blocking-GC.md
wy65701436: actually, the step 1 is enough for Mark, all of referenced blobs are in table project&blob.
But add step 2 is just double guarantee that any of referenced blob by artifact will not be in the candidate.

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+#### Question 3, how to deal with the uploading "untagged manifest" of a index at the phase of sweeping.
+Different wit push a stand alone artifact, docker client will send a head request before putting a manifest in push a **untagged** manifest in the process of pushing Index,  we will intercept that request.
+
+![head_manifest](../images/non-blocking-gc/head_manifest.png)
+
+#### Question 4, what about if client only sends head request, and no put following.
+The manifest or blob will only be removed from the deletion candidates, and it can be GCed in the next execution.
+
+### Delete Blob & Manifest
+We'd like to enable the registry controller to have the capability to delete blob & manifest via digest by leveraging the distribution code as library.
```
reasonerjt: I think it only needs to support deleting blobs?  Do we need to treat manifest any differently?

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+
+**DOUBLE GUARANTEE**
+
+We need to introduce the cutoff time, any to be deleted blob & manifest, the update time must not be later than the cutoff time.
```
reasonerjt: Is there update time of a blob in DB?

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+
+```
+
+Draft code PR has been created: https://github.com/goharbor/harbor/pull/10441
+
+### Overall flow
+The basic flow is:
+
+* Mark the GC candidates in Harbor Core
+* Trigger a GC job and pass the candidates.
```
reasonerjt: Any reason not to do the marking in GC job?

In GC job you will remove untagged artifacts, and that will increate the set of candidiates?

proposals/new/Non-Blocking-GC.md
wy65701436: It's a new column of table blob.

1， all of blobs in candate are `delete` staus
2,   sweeper pick up one blob, and mark it as `deleting`, then to remove it.
3,   eventually, the blob will be remove from table blob.

Maybe, I should add this into proposal.

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+We do have a table to record the uploading blobs info, that's project & blob.
+The delete candidate excludes all of blobs that in the project & blob.
+
+1. candidate set 1 -- all blobs from table blob exclude the items in the table project & blob.
+2. candidate set 1 excludes all of referenced blobs (artifact -> artifact & blob -> blob).
+
+![mark_uploading](../images/non-blocking-gc/mark_uploading.png)
+
+### Sweep
+The registry controller will grant the capability of deleting blob & manifest.
```
reasonerjt: I don't quite follow the questions and solutions in this section.

Generally I don't quite understand the deleting and delete status, could you explain what are they and how do you check it?

And I'm not very sure it's safe to assume the `HEAD` request.  What if there's a 3rd party client that doesn't do the HEAD before pushing?

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Question 3, how to deal with the uploading "untagged manifest" of a index at the phase of sweeping.
+Different wit push a stand alone artifact, docker client will send a head request before putting a manifest in push a **untagged** manifest in the process of pushing Index,  we will intercept that request.
+
+![head_manifest](../images/non-blocking-gc/head_manifest.png)
+
+#### Question 4, what about if client only sends head request, and no put following.
+The manifest or blob will only be removed from the deletion candidates, and it can be GCed in the next execution.
+
+### Delete Blob & Manifest
```
reasonerjt: In theory, there can race condition two processes (registry/ registry ctl) calling the storage API to handle same blob, do we assume the storage should handle such situation?
Let's find some proof this assumption is reliable, and reference it in this proposal.

proposals/new/Non-Blocking-GC.md
```diff
@@ -0,0 +1,297 @@
...
+
+
+### Mark
+
+Bases on the Harbor DB, we can count each blob/manifest's reference count, and select the reference count 0 as the candidate.
+![mark](../images/non-blocking-gc/mark.png)
+
+
+#### Question 1, how to deal with the uploading blobs at the phase of marking.
+We do have a table to record the uploading blobs info, that's project & blob.
```
mmpei, line 143: it makes things too heavy. and concurrent risk still exists.

mmpei:
it brings complex when you decide to treat Image GC as a immediate job. Nobody can avoid 'Stop The World'. Maybe we can treat it as an offline job and users only want to delete the blobs that has been stale for a long time but not the blobs in active.

steven-zou:

PR #143, opened 2020-05-31
Proposal for enhanced default processor to handle user-defined artifact like Machine Learning artifacts

hyy0322:
Signed-off-by: Yiyang Huang huangyiyang.huangyy@bytedance.com


gaocegege:
Proposal Preview: [proposals/artifact-processor-extender.md](https://github.com/hyy0322/community/blob/artifact-processor-extender/proposals/artifact-processor-extender.md)

/cc @zhujian7 @gaocegege @hainingzhang 

/assign @xaleeks @steven-zou

reasonerjt:
Hi,

Thanks for the proposal.

I find the framework proposed here may be a bit over-engineering at least for current requirements.  It also introduces new attack-surface, for example the robot accounts in the external processors may be leaked and a malicious user can use it to pull artifacts from Harbor.

If we define a schema in the config blob of an artifact, it seems the processor extender is not very necessary?

i.e. if a in-tree processor and extract data as a re-defined contract the artifact developer will only have to write the config JSON object in a certain format, then he will not need to implement his processor and we don't need to introduce yet another framework to manage external service registration in Harbor?

gaocegege:
Currently, it may work. In our use case, we now just have one possible addition: README.

In the future, we will support a new type `ML Dataset`. And we will keep a tiny part of the dataset in one blob layer as one addition. It will be used to show the directory struture of the dataset. These kinds of artifacts require higher extensibility, I think.

We can have the detailed dicusstions in the community meeting today :beers: We are open to the decision. If the community members think it is too far, we can keep a in-tree common processor  instead.

gaocegege:
@reasonerjt After a detailed review, I think in-tree processor is also one possible approach. We will refine the proposal soon.

proposals/artifact-processor-extender.md
```diff
@@ -0,0 +1,579 @@
...
+}
+```
+
+<p align="center">
+<img src="images/artifact-processor-extender/in-tree-workflow.png" height="600">
+<p align="center">Fig. 4 Workflow of Pushing an Artifact using the General Processor</p>
+</p>
+
+
+## Out-of-tree processor
```
xaleeks, line 299: get rid of this section?  I think we all agreed on in-tree approach yes?

gaocegege: I think we can keep but rename the section to deprecated option. 

Because maybe we will move to it in the future. WDYT

proposals/artifact-processor-extender.md
```diff
@@ -0,0 +1,592 @@
...
+```
+As is showed above, we can define a media type ```application/vnd.custom.artifact.config.v1+json``` to show it is a media type of user-defined artifact config
+
+
+config layer
+```
+{ 
+    // user defined config
+    // abstract metadata will use this config data as extra attrs
+    "custom": {},
```
gaocegege, line 198: Prefer not to have such a "custom" object, we can embed user config directly in root like this:

```diff
{
+ "harbor": {
+        // artifact config
+        "artifactMetadata":{
+            // artifact type
+            "type": "ORMB",
+            // artifact media type
+            "mediaType": "application/vnd.caicloud.ormb.xxx",
+            // additions
+           "additions": [
+                {
+                    // content type
+                    "contentType": "plain"
+                    // addition type name, one of addition type
+                    "type": "type_xxx",
+                    "digest": "sha256:xxx"
+                },
+                {
+                    "contentType": "json",
+                    "type": "type_xxx",
+                    "digest": "sha256:xxx"
+                },
+                {
+                    "contentType": "markdown"
+                    "type": "type_xxx",
+                    "digest": "sha256:xxx"
+                },
+                ...
+            ]
+        }
+    }
   "created": "2015-10-31T22:22:56.015925234Z",
   "author": "Ce Gao <gaoce@caicloud.io>",
   "description": "CNN Model",
   "tags": [
       "cv"
   ],
   "labels": {
       "tensorflow.version": "2.0.0"
   },
   "framework": "TensorFlow",
   "format": "SavedModel",
   "size": 9223372036854775807,
   "metrics": [
       {
           "name": "acc",
           "value": "0.9"
       }
   ],
   "hyperparameters": [
       {
           "name": "batch_size",
           "value": "32"
       }
   ],
   "signature": {
       "inputs": [
           {
               "name": "input_1",
               "size": [
                   224,
                   224,
                   3
               ],
               "dtype": "float64",
           }
       ],
       "outputs": [
           {
               "name": "output_1",
               "size": [
                   1,
                   1000
               ],
               "dtype": "float64",
           }
       ],
       "layers": [
           {
               "name": "conv"
           }
       ]
   },
   "training": {
       "git": {
           "repository": "git@github.com:caicloud/ormb.git",
           "revision": "22f1d8406d464b0c0874075539c1f2e96c253775"
       }
   },
   "dataset": {
       "git": {
           "repository": "git@github.com:caicloud/ormb.git",
           "revision": "22f1d8406d464b0c0874075539c1f2e96c253775"
       }
   }
}
```

proposals/artifact-processor-extender.md
```diff
@@ -0,0 +1,592 @@
...
+Links:
+
+- Discussion: [goharbor/harbor#12013](https://github.com/goharbor/harbor/issues/12013)
+- Slides: [Feature Request: Harbor Artifact Processor Extender](https://docs.google.com/presentation/d/1rX7v9IhXjXEAmbJP29nkSU2RJXUrpLJujL8iemqcjnU/edit#)
+
+## Abstract
+
+Harbor v2.0 makes Harbor the first OCI-compliant open source registry capable of storing a multitude of cloud-native artifacts like container images, Helm charts, OPAs, Singularity, and much more. It found strong demand on extending artifact types to support more complex scenarios. But the artifact authors now have to implement the processing logic in Harbor Core, which is not extensible.
+
+The current design might go against the adoption of Harbor in industries since there are always proprietary artifact types. Thus we design the artifact processor extender to address the problem in this proposal. When new custom artifact types are registered into Harbor Core, the core service will communicate with the remote artifact processor extender via RESTful API. Artifact-specific logic will keep in the extender, and the Harbor Core will be artifact-neutral.
```
zhujian7, line 49: I think the description of `the core service will communicate with the remote artifact processor extender via RESTful API` is not accurate. Only the `out-of-tree` approach needs `RESTful API`?

proposals/artifact-processor-extender.md
hyy0322: I agree with it.

proposals/artifact-processor-extender.md
hyy0322: It is correct, ```remote artifact processor extender``` means using out-of-tree approach

proposals/artifact-processor-extender.md
```diff
@@ -0,0 +1,594 @@
...
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	AbstractAddition(ctx context.Context, artifact *artifact.Artifact, additionType string) (addition *Addition, err error)
+}
+```
+
+#### ```GeneralProcessor``` Implement
+```
+func (g *GeneralProcessor) GetArtifactType(artifact *artifact.Artifact) string {
+	configLayer := pullBlob(artifact.manifest.config.digest)
+	return configLayer.Type
```
zhujian7, line 279: `configLayer.Type` -> `configLayer.harbor.artifactMetadata.type` ?

proposals/artifact-processor-extender.md
hyy0322: It cames to me if we embed user config directly in root, it is difficult to ```AbstractMetadata```. Because we need to put user config in artifact.extraattrs.
```
func (g *GeneralProcessor) AbstractMetadata(artifact *artifact.Artifact) error {
	configLayer := pullBlob(artifact.manifest.config.digest)
	artifact.ExtraAttrs = configLayer.custom
}
```

gaocegege: We can get all config into it.

hyy0322: Oh yes, we can do it this way. I know how to do it.

proposals/artifact-processor-extender.md
hyy0322: Done.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,51 @@
...
+    "$id": "https://raw.githubusercontent.com/goharbor/community/master/proposals/assets/artifact-processor/artifact-config-layer-schema.json",
+    "type": "object",
+    "title": "Artifact config layer schema",
+    "description": "The schema comprises the entire configuration JSON document in artifact config layer",
+    "default": {},
+    "examples": [
+        {
+            "key1": "value1",
+            "key2": "value2",
+            "registryHarbor": {
```
gaocegege, line 12: Prefer registry/harbor personally.

gaocegege:
@xaleeks @reasonerjt @zhujian7 @steven-zou 

We updated the proposal recently. I'd appreciate it if you could have another review.

Thanks.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: I think use registry/harbor in json is wired.

gaocegege: Since [JSON LD](https://en.wikipedia.org/wiki/JSON-LD) uses @ in JSON key, I think / also works. But I have no objection to it.

steven-zou: So `X-Harbor-Attributes` do not match the key convention?

steven-zou: I agree `registry/harbor` is not a good term.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+
+- 2020-06-24 Draft v3
+- 2020-06-15 [Draft v2](https://github.com/hyy0322/community/blob/7349e0a4325d021b9d52ed61afbc6118f30c7774/proposals/artifact-processor-extender.md)
+- 2020-05-28 [Draft v1](https://github.com/gaocegege/community-2/blob/enhancement/proposals/artifact-processor-extender.md)
+
+## Abstract
+
+Harbor v2.0 makes Harbor the first OCI-compliant open source registry capable of storing a multitude of cloud-native artifacts like container images, Helm charts, OPAs, Singularity, and much more. It found strong demand on extending artifact types to support more complex scenarios. But the artifact authors now have to implement the processing logic in Harbor Core, which is not extensible.
+
+The current design might go against the adoption of Harbor in industries since there are always proprietary artifact types. Thus we design the artifact processor extender to address the problem in this proposal. When new custom artifact types are registered into Harbor Core, the core service will communicate with the remote artifact processor extender via RESTful API. Artifact-specific logic will keep in the extender, and the Harbor Core will be artifact-neutral.
```
steven-zou, line 40: `via RESTful API`?  I think this is a mistake.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+  "type": "CHART"
+}
+```
+
+The self-contained response is also necessary for these user-defined artifact types. Or we cannot use Harbor directly for most scenarios. The `extra_attrs` field is processed by the Helm Chart processor, which is an implementation of artifact processor interface `Processor`.
+
+The current design of the artifact processor is shown in Fig. 1. `Processor` interface is defined in Harbor Core, and there are four implementations for different types which embeds `base.IndexProcessor` and `base.ManifestProcessor`.
+
+<p align="center">
+<img src="images/artifact-processor-extender/current-design.png" height="330">
```
steven-zou, line 125:
```suggestion
<img src="images/artifact-processor/current-design.png" height="330">
```

All the other image links have the similar issues.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+<p align="center">Fig. 1 Current Design of Harbor Artifact Processor</p>
```
steven-zou, line 126: Picture content is out of date? (related to HTTP REST approach)

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+This proposal is to:
+
+- Define Harbor-specific json schema to make default processor able to process user defined artifact.
+- Keep non-invasive to the current built-in processors, at the same time.
+
+## Non-Goals
+
+This proposal is not to:
+
+- Support whitelist for artifact types. [goharbor/harbor#12061](https://github.com/goharbor/harbor/issues/12061)
```
steven-zou, line 154:
```suggestion
- Support allowlist for artifact types. [goharbor/harbor#12061](https://github.com/goharbor/harbor/issues/12061)
```

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+
+    // harbor defined config
+    "registryHarbor": {
+        "schemaVersion": 1,
+        "icon": "https://github.com/caicloud/ormb/raw/master/docs/images/logo.png",
+        // additions
+        "additions": [
+            {
+                // content type
+                "contentType": "plain",
```
steven-zou, line 185: Use standard mime types? application/**?

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,298 @@
...
+		Content: additionLayer,
+		ContentType: additionType,
+	}, nil
+}
+```
+
+<p align="center">
+<img src="images/artifact-processor-extender/in-tree-workflow.png" height="600">
+<p align="center">Fig. 4 Workflow of Pushing an Artifact using the Default Processor</p>
+</p>
```
steven-zou, line 298: Please add a development plan here to describe the draft plan for delivering such a feature. Thanks

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: According to [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml)
- Property names should be meaningful names with defined semantics.
- Property names must be camel-cased, ascii strings.
- The first character must be a letter, an underscore (_) or a dollar sign ($).
- Subsequent characters can be a letter, a digit, an underscore, or a dollar sign.
- Reserved JavaScript keywords should be avoided (A list of reserved JavaScript keywords can be found below).

 ```xHarborAttributes``` or ```x_harbor_attributes``` is a good term. Camel case is widely used. I think ```registryHarbor``` and ```xHarborAttributes``` both are ok. I'm open to it.

proposals/enhanced-default-processor.md
hyy0322: Picture content just point out current design in Harbor core. It is not out if date. HTTP REST approach has been removed.

hyy0322:
> Please fix the broken links first.

Sorry, I have updated the links.

proposals/enhanced-default-processor.md
hyy0322: Development plan added

proposals/enhanced-default-processor.md
hyy0322: Ok, ```text/plain``` will be used.

proposals/enhanced-default-processor.md
hyy0322: I have wrote about this content.

hyy0322:
@steven-zou Thanks for review. Made some low-level mistakes.

renmaosheng:
I have one comment regarding the artifact you plan to define and implement, could the new type be used in common kubeflow use case?

hyy0322:
> I have one comment regarding the artifact you plan to define and implement, could the new type be used in common kubeflow use case?

Hi, @renmaosheng. I don't know if I understand your question right.

You mean ormb artifact ?

We are not going to define a new type of artifact in this proposal. We design a specification in artifact config layer. If users want to show their user-defined artifact data, maybe config kv as extra_attrs or specific layer data as additon in Harbor, they must follow the specification so that Harbor-core could recognize it. Artifact is artifact itself. I think it is not related with usage in kubeflow. @gaocegege Am I right?

hyy0322:
@xaleeks @reasonerjt  @steven-zou @zhujian7 @gaocegege 

PTAL. Thanks!

renmaosheng:
> > I have one comment regarding the artifact you plan to define and implement, could the new type be used in common kubeflow use case?
> 
> Hi, @renmaosheng. I don't know if I understand your question right.
> 
> You mean ormb artifact ?
> 
> We are not going to define a new type of artifact in this proposal. We design a specification in artifact config layer. If users want to show their user-defined artifact data, maybe config kv as extra_attrs or specific layer data as additon in Harbor, they must follow the specification so that Harbor-core could recognize it. Artifact is artifact itself. I think it is not related with usage in kubeflow. @gaocegege Am I right?
Thanks Yiyang. I understood that your will not change the artifact type in this proposal. But I am more interested in your own artifact defined, if it is commonly used by kubeflow, maybe in future that type can be a new type supported in Harbor like OPA bundle,etc.

gaocegege:
@renmaosheng 

> Thanks Yiyang. I understood that your will not change the artifact type in this proposal. But I am more interested in your own artifact defined, if it is commonly used by kubeflow, maybe in future that type can be a new type supported in Harbor like OPA bundle,etc.

Hi, Maosheng. If you are interested in our open-sourced artifact type for Machine Learning models, please have a look at https://github.com/caicloud/ormb 

It is not a project under Kubeflow organization now but we will investigate if we can (or should) contribute it to Kubeflow.

hyy0322:
@steven-zou @reasonerjt Hello, could you please have another review. Thanks a lot!

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: I agree.

hyy0322: > I preferred the `xHarborAttributes` as I can understand what `xHarborAttributes` means without any thinking, but `registryHarbor` can not. And `xHarborAttributes` can prevent conflicts with user-defined attributes.

I agree.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,51 @@
...
+            "title": "User defined config key value",
+            "description": "key1 can be any string, it is user defined attribute about artifact, currently it only support string type",
+            "default": "",
+            "examples": [
+                "value1"
+            ]
+        },
+        "registryHarbor": {
+            "description": "The schema comprises the Harbor-specific configuration JSON document in artifact config layer",
+            "$ref": "https://raw.githubusercontent.com/goharbor/community/master/proposals/assets/artifact-processor/artifact-config-layer-schema.json"
```
zhujian7, line 48: I think the ref should be `https://raw.githubusercontent.com/goharbor/community/master/proposals/assets/artifact-processor/registry-harbor-schema.json`.  Please check.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,51 @@
...
+                "denyKeyList": [
+                    "key1"
+                ]
+            }
+        }
+    ],
+    "required": [],
+    "additionalProperties": true,
+    "properties": {
+        "key1": {
```
zhujian7, line 36: I preferred to put the `key1` in the example field instead of here.

zhujian7:
Should we explain the rules of parsing `artifact type` here to indicate users must conform to the OCI manifest config mediaType?

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: https://github.com/opencontainers/image-spec/blob/master/descriptor.md
mediaType already have convention.

hyy0322:
> Should we explain the rules of parsing `artifact type` here to indicate users must conform to the OCI manifest config mediaType?

https://github.com/opencontainers/image-spec/blob/master/descriptor.md
mediaType already have convention.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: I will remove it

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: OK

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,103 @@
...
+                    {
+                        "contentType": "text/plain",
+                        "name": "readme",
+                        "digest": "sha256:xxx"
+                    }
+                ],
+                "denyKeyList": [
+                    "metrics"
+                ]
+            },
```
hyy0322, line 28: In last online meeting, we disscused that we add key ```extraAttributes``` in ```xHarborAttributes```, like
```
"xHarborAttributes": {
    "schemaVersion": 1,
    "icon": "https://github.com/caicloud/ormb/raw/master/docs/images/logo.png",
    "additions": [
	    {
	        "contentType": "text/plain",
	        "name": "yaml",
	        "digest": "sha256:xxx"
	    },
	    {
	        "contentType": "text/plain",
	        "name": "readme",
	        "digest": "sha256:xxx"
	    }
	],
	"extraAttributes": {
             "denyKeyList": [
        	    "metrics"
    	     ]
	}
}
```
I don't think we need this key any more. All keys can add in ```xHarborAttributes```.  
WDYT @gaocegege @reasonerjt @steven-zou

hyy0322:
In last online meeting, we disscussed that we parse the type of value of user's key in string type. I read source code again, I think we don't need to limit to parse the value type to string. Just like manifest processor, use ```interface``` to parse any type is ok. WDYT @steven-zou 
https://github.com/goharbor/harbor/blob/master/src/controller/artifact/processor/base/manifest.go#L55-L71

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,361 @@
...
+	// try to parse the type from the media type
+	strs := artifactTypeRegExp.FindStringSubmatch(d.mediaType)
+	if len(strs) == 2 {
+		return strings.ToUpper(strs[1])
+	}
+	// can not get the artifact type from the media type, return unknown
+	return ArtifactTypeUnknown
+}
+
+func (d *defaultProcessor) ListAdditionTypes(ctx context.Context, artifact *artifact.Artifact) []string {
```
ywk253100, line 302: As we'll store config layer as extra attributes in database and the extra attributes contain the information of additions, when `ListAdditionTypes(ctx context.Context, artifact *artifact.Artifact)`  is called, the extra attributes should already be read and populated into the input parameter `artifact`, we can get the additions directly from it and don't need to read the config layer from storage again

hyy0322: Yes. Thanks. I will correct it.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: /cc @ywk253100 WDYT

hyy0322:
> In last online meeting, we disscussed that we parse the type of value of user's key in string type. I read source code again, I think we don't need to limit to parse the value type to string. Just like manifest processor, use `interface` to parse any type is ok. cnab, chart use manifest processor to populate metadata. And in our ormb artifact, the value of user key have some none string type. WDYT @steven-zou
> https://github.com/goharbor/harbor/blob/master/src/controller/artifact/processor/base/manifest.go#L55-L71

@ywk253100 WDYT

proposals/assets/artifact-processor/artifact-config-layer-schema.json
ywk253100: I agree. All keys can put under `xHarborAttributes` directly

ywk253100:
> > In last online meeting, we disscussed that we parse the type of value of user's key in string type. I read source code again, I think we don't need to limit to parse the value type to string. Just like manifest processor, use `interface` to parse any type is ok. cnab, chart use manifest processor to populate metadata. And in our ormb artifact, the value of user key have some none string type. WDYT @steven-zou
> > https://github.com/goharbor/harbor/blob/master/src/controller/artifact/processor/base/manifest.go#L55-L71
> 
> @ywk253100 WDYT

`interface` is more flexible

hyy0322:
> > > In last online meeting, we disscussed that we parse the type of value of user's key in string type. I read source code again, I think we don't need to limit to parse the value type to string. Just like manifest processor, use `interface` to parse any type is ok. cnab, chart use manifest processor to populate metadata. And in our ormb artifact, the value of user key have some none string type. WDYT @steven-zou
> > > https://github.com/goharbor/harbor/blob/master/src/controller/artifact/processor/base/manifest.go#L55-L71
> > 
> > 
> > @ywk253100 WDYT
> 
> `interface` is more flexible

ok

proposals/assets/artifact-processor/artifact-config-layer-schema.json
gaocegege: LGTM

steven-zou: +1

steven-zou: Let's clarify again here, `denyKeyList` is used to tell the processor which keys can be ignored and no need to extracted out.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,360 @@
...
+	return &Addition{
+		Content: additionLayer,
+		ContentType: additionType,
+	}, nil
+}
+```
+
+<p align="center">
+<img src="images/artifact-processor/in-tree-workflow.png" height="600">
+<p align="center">Fig. 4 Workflow of Pushing an Artifact using the Default Processor</p>
```
steven-zou, line 356: I‘m not sure whether I correctly understand the flow or not. Just clarify here, the default processor will be used only when the uploading artifact has no assigned processors (no effect to image, image index, CNAB etc.), right?

When the default processor processing the config.json, we'll check if it matches the harbor-specified schema defined in this proposal. If yes, then continue the data processing procedure, if not, do nothing or?

@hyy0322 @gaocegege

steven-zou: @ywk253100 What's your opinion?

gaocegege: > When the default processor processing the config.json, we'll check if it matches the harbor-specified schema defined in this proposal. If yes, then continue the data processing procedure, if not, do nothing

I think it is the workflow we want.

hyy0322: > I‘m not sure whether I correctly understand the flow or not. Just clarify here, the default processor will be used only when the uploading artifact has no assigned processors (no effect to image, image index, CNAB etc.), right?
> 
> When the default processor processing the config.json, we'll check if it matches the harbor-specified schema defined in this proposal. If yes, then continue the data processing procedure, if not, do nothing or?
> 
> @hyy0322 @gaocegege

Yes, we will do some check. 
1. check ```xHarborAttibutes``` key exist or not.
2. check schema version.
After these check passed, start to process. Or do nothing.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: > Let's clarify again here, `denyKeyList` is used to tell the processor which keys can be ignored and no need to extracted out.

ok

xaleeks: just curious, what was the original intent of the extraAttributes?

and what is the usefulness of denyKeyList?

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,103 @@
...
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://raw.githubusercontent.com/goharbor/community/master/proposals/assets/artifact-processor/artifact-config-layer-schema.json",
+    "type": "object",
+    "title": "Artifact config layer schema",
+    "description": "The schema comprises the entire configuration JSON document in artifact config layer",
+    "default": {},
+    "examples": [
+        {
+            "xHarborAttributes": {
+                "schemaVersion": 1,
```
reasonerjt, line 11: To better illustrate it's the very first version and immature, can we make it 0.1?

proposals/assets/artifact-processor/artifact-config-layer-schema.json
```diff
@@ -0,0 +1,103 @@
...
+                        "name": "yaml",
+                        "digest": "sha256:xxx"
+                    },
+                    {
+                        "contentType": "text/plain",
+                        "name": "readme",
+                        "digest": "sha256:xxx"
+                    }
+                ],
+                "denyKeyList": [
```
reasonerjt, line 25: How about replacing `deny` with `skip`?

proposals/assets/artifact-processor/artifact-config-layer-schema.json
reasonerjt: +1 we should clarify the usage and how the `extra_attributes` are extracted

hyy0322: Ok, I will add more explanation about this part.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: Yes, I think ```skip``` is better.

proposals/assets/artifact-processor/artifact-config-layer-schema.json
hyy0322: This is a image manifest
```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   ...
   "layers": []
}
```
@reasonerjt I'v considered about of use 0.1, but image manifest use int instead of float number. Is it ok we use 0.1? If ok, I think 0.1 is better.

gaocegege: Ref https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions

Personally, prefer keeping consistency with the image manifest config to avoid confusing readers/users.

hyy0322: > Ref https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions
> 
> Personally, prefer keeping consistency with the image manifest config to avoid confusing readers/users.

Yes, this is my concern about it.

steven-zou: I think 1 is ok

hyy0322: So, @reasonerjt could we all agree on using int type 1?

reasonerjt: I don't understand why it has to be consistent with the manifest in OCI spec.  
The version is for Harbor's attributes and it has its own versioning.  The reason I prefer 0.1 is that it reflects its immaturity and implies it may subject to change.

gaocegege: I do not think 0.1 implies that it may subject to change. Is there any mechanism in Harbor to allocate that the faeture is alpha/beta or experimental? I think we should notify users somewhere instead of implying.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,268 @@
...
+
+```json
+{
+    "schemaVersion": 2,
+    "config": {
+        "mediaType": "application/vnd.caicloud.model.config.v1alpha1+json",
+        "digest": "sha256:be948daf0e22f264ea70b713ea0db35050ae659c185706aa2fad74834455fe8c",
+        "size": 187,
+        "annotations": {
+            "io.goharbor.artifact.schema.version": "0.1",
```
steven-zou, line 194:
```suggestion
            "io.goharbor.artifact.schema.version": "v1/alpha",
```

steven-zou, line 194: I'm not sure if `io` is easy for understanding. I'm thinking maybe org.goharbor is also ok as goharbor is an OSS community org.

gaocegege: SGTM

gaocegege: Addressed

gaocegege: Addressed

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,268 @@
...
+- The enhanced default processor implementation to support user-defined artifacts.
+
+### Harbor-specific Configuration
+
+We introduce a harbor-specific configuration in the manifest annotations, which contains information that Harbor core needs to understand the artifact.
+
+The harbor-specific configuration follows the style of [OCI Pre-Defined Annotation Keys](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys). There are two proposed keys in `manifest.config.annotations`:
+
+- **org.goharbor.artifact.skiplist** The list of skip keys. Harbor will ignore these keys in configuration.
+- **org.goharbor.artifact.schema.version** The schema version.
```
reasonerjt, line 178: instead of introducing a key for schema version, would it be better to change the style to 
`artifact.config.goharbor.io/v1schema1/skiplist`

such that each annotation has a version, including the ones that attach to layers?

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,268 @@
...
+- The Harbor-specific schema in artifact manifest to tell Harbor core more information about the user-defined artifacts.
+- The enhanced default processor implementation to support user-defined artifacts.
+
+### Harbor-specific Configuration
+
+We introduce a harbor-specific configuration in the manifest annotations, which contains information that Harbor core needs to understand the artifact.
+
+The harbor-specific configuration follows the style of [OCI Pre-Defined Annotation Keys](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys). There are two proposed keys in `manifest.config.annotations`:
+
+- **org.goharbor.artifact.skiplist** The list of skip keys. Harbor will ignore these keys in configuration.
```
reasonerjt, line 177: Please also clarify the delimeter of the keys

hyy0322: ok

proposals/enhanced-default-processor.md
hyy0322: I think version info in annotation key is wired. We can specify which keys are supported in certain verison. And organization keys in code.

hyy0322: For example, let me show you demo.
image

hyy0322: Version info in key, if we upgrade version, all keys need to change. Or we just need to change value for annotation schema version key.

ywk253100: I personally prefer the suggestion of @reasonerjt that make the schema version as part of each annotations, it's more strict and clear for users when trying to build their artifacts.

Think about that when a user tries to make the blob as the icon, he just needs to add an annotation in the blob. If we put the version in a separated annotation, he must add the version annotation first, it isn't very straightforward. And where we put the version annotation in? In the annotation of config or manifest? Both of them are weird.

I suggest the format of annotation as follows:

`io.goharbor.v1alpha1.artifact.config.skiplist`
`io.goharbor.v1alpha1.artifact.icon`

The `io.goharbor` is the reverse domain of Harbor.

The `v1alpha1` is the version of annotation schema, we can use a single one version for all of the annotations and I suggest we just maintain one specific version of annotation in one specific release.

The `config.skiplist` and `icon` is the concrete annotation keys. As the `skiplist` indicates the skip list for config items, a `config` prefix is more reasonable.

gaocegege: https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

You can see that there is no version info in the OCI predefined annotation key. From the perspective of the artifact vendor, I do not like this approach since we have to take care of version for every key.

gaocegege: And, as a vendor, I do not think I will update the version for just one key. I cannot figure out such a use case. As you know, artifact also has its version mechanism. We do not need so much flexibility.

proposals/enhanced-default-processor.md
```diff
@@ -0,0 +1,268 @@
...
+            "org.goharbor.artifact.skiplist": "metrics,git"
+        }
+    },
+    "layers": [
+        {
+            "mediaType": "image/png",
+            "digest": "sha256:d923b93eadde0af5c639a972710a4d919066aba5d0dfbf4b9385099f70272da0",
+            "size": 166015,
+            "annotations": {
+                "org.goharbor.artifact.icon": "true"
```
ywk253100, line 204: Seems the value of the icon annotation `true` is meaningless. What will happen if set it to `false`?
How about empty it or put the format(png, jpeg, ...) in it?

gaocegege: Format is shown in mediaType. We can set it empty, I think.

hyy0322: I think we can empty it. We just need key ```org.goharbor.artifact.icon```.

steven-zou:
```
io.goharbor.artifact.v1alpha1.icon
io.goharbor.artifact.v1alpha1.skip-list
```

gaocegege:
The PR is merged, please update the proposal according to our implementation @hyy0322 

👍

hyy0322:
Proposal updated.

steven-zou (approved):
approved

steven-zou:

PR #145, opened 2020-06-24
Enhanced data  schema for Cloud-native Workload Vulnerability Data Storage for Analytics, Reporting and Policy Evaluation.

prahaladdarkin:
This proposal will lay the foundational work that allow advanced reporting, analytics, policy evaluation and integration with third-party systems providing these capabilities such as Open Policy Agent.
The contents of this proposal are based on the document describing the Cloud Native Workload - Common Vulnerability Evaluation and Reporting Framework.


proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+As can be seen above the `Report` data structure no longer holds a JSON blob for the entire report. Each record is represented by the `VulnerabilityRecord` structure shown below
+
+### Create VulnerabilityReport data structure mapped to vulnerability_data table
+```
+// VulnerabilityRecord identifies an  individual vulnerability item in the scan.
+// Identified by the `cve_id`.
+// Relates to the image using the `digest` and to the report using the `report UUID` field
+type VulnerabilityRecord struct {
```
reasonerjt, line 93: There may be duplicated records with same CVE ID, only different ID and report_uuid.

prahaladdarkin: I have modified the schema to include a junction table that will store the relation between the Vulnerability Records and the reports. Please take a look. 
I would want to know whether ReportUUID would be a good foreign key in the junction table or should it be the auto-generated ID.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+	VulnRecord int64  `orm:"column(vuln_record_id)`
+}
+```
+
+### Example Ad-hoc queries
+The types of ad-hoc queries that can be made possible using the above schema are illustrated below
+
+### Number of software artifacts with a particular CVE-ID across all projects
+
+`select count(1) from vulnerability_record where CVEID = '<CVE_ID>'`
```
danielpacak, line 130: It's tempting to have a UI for admins to search for packages with the specified vulnerability. Do we plan to leverage the proposed schema to immediately add simple search / reporting capabilities to the Harbor interface?

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+* OS and software package related vulnerabities
+* License related vulnerabiitlies will be raised as a spearate proposal since it also requires integrating Harbor with license scanning tools like [Tern](https://www.linuxfoundation.org/blog/2019/12/tern-1-0-0-is-generally-available/)
+
+## Image OS and package level Vulnerabilities
+
+These represent the vulnerabilities within an OS image or software package. The schema to describe such vulnerabilities is presented below. This schema is builds upon the [Pluggable scanner spec](https://github.com/goharbor/pluggable-scanner-spec) and introduces the following new fields
+
+|Attribute|Description|
+| --- | --- |
+|CVE_3_Score| The score in the CVSS 3.0 format that can be associated with the image|
```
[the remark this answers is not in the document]
prahaladdarkin: Excellent comment and queries related to CVSS.
Right now this column represents the overall Qualitative score across all vectors. Given the related comment about CVSS vectors being part of the schema, it would probably make sense to do the following:
* This table maintains the overall qualitative score of the image CVSS2 or CVSS3. 
* A CVSS 3.0 detail schema table can be created where-in we store the data about the attack vectors and also the CIA score for each. 
* Store a reference to the CVSS detail schema rows within this table. This can be null to accommodate scanners that do not provide this information.

Additionally, do we need to worry about CVSS 2.0? As far as I understand this standard is slowly being phased out in favor of CVSS 3.0 which provides more standardized results. In the original document to Liz Rice had commented about the aspect of considering only CVSS 3.0. I am not sure if the default Claire scanner provides CVSS 3.0 score.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+|VendorAttributes|Additional **key-value** pairs containing vendor specific attributes that provide more context around the vulnerability. It is not necessary that this structure is present every time or for every vendor. Implementations are free to populate relevant data within this data structure to capture vulnerability specific information.|
+
+
+## Enhancements to the current data storage mechanisms
+ 
+The scan report produced vulnerability scanners in Harbor are stored within a JSON blob within the Harbor DB. This prevents ad-hoc querying of the vulnerability data to provide more richer reporting and analytics
```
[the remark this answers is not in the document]
prahaladdarkin: Done

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+// Relates to the image using the `digest` and to the report using the `report UUID` field
+type VulnerabilityRecord struct {
+	ID               int64    `orm:"pk;auto;column(id)"`
+	CVEID            string   `orm:"unique;column(cve_id)"`
+	Digest           string   `orm:"column(digest)"`
+	Report           string   `orm:"column(report_uuid)"`
+	Package          string   `orm:"column(package)"`
+	PackageType      string   `orm:"column(package_type)"`
+	Severity         string   `orm:"column(severity)"`
+	Fix              string   `orm:"column(fixed_version)"`
```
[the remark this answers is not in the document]
prahaladdarkin: Added. I have also provided an example in the attached document in response to the comments related to package version. 
An additional comment I added and require views on is whether the PackageType (OS/SoftwareLib) be actually deduced by the scanner. While I did introduce the field in the original document, in retrospect I do see that the field is redundant and does not provide any valuable information.

proposals/new/common_vulnerability_schema_proposal.md
prahaladdarkin: @xaleeks  and @steven-zou  can comment. IMO the immediate advantage of this schema should be leveraged. A set of pre-canned reports that satisfy the most pressing reporting use cases can be shipped with Harbor.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+* Allow ad-hoc reporting and analytics on the Image vulnerability data
+* Author and share standard reporting queries to fetch image vulnerability data
+* Author and share policies (OPA or custom) that work in conjunction with the enhanced data schema to report on the policy compliance posture of images. 
+* Gain deep insights into all aspects of the image compliance posture (licenses,  image file system structure, file content) in addition to traditional OS and software package level vulnerabilities.
+* Facilitate storage of standard vulnerability information as well as vendor specific additional information/attributes
+
+
+## Discussions and related work
+
+* [Cloud Native Workload - Common Vulnerability Evaluation and Reporting Framework](#)
```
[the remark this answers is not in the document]
prahaladdarkin: Attached. Had originally sent it over email

proposals/new/common_vulnerability_schema_proposal.md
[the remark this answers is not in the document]
prahaladdarkin: @danielpacak  I have introduced CVSS related schema definitions now as a response to this comment. Please take a look

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,157 @@
...
+//ReportVulnerabilityRecord defines the relation between a vulnerability record
+// and the report containing the vulnerability record.
+//This junction table optimizes storage space usage since a vulnerability can be
+//a member of many reports and the information about the vulnerability need not be
+//repeated.
+//Identified by Report.UUID and the VulnRecord.Id
+type ReportVulnerabilityRecord struct {
+	ID         int64  `orm:"pk;auto;column(id)"`
+	Report     string `orm:"column(report_uuid)"`
+	VulnRecord int64  `orm:"column(vuln_record_id)`
```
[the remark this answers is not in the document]
prahaladdarkin: Currently, I observed that two scans for the same image digests generate different scan report UUIDs. IMO the vulnerability record should have a version field which would be updated each time there is an update to the data from the upstream database. The table here would have a composite key of Report, VulnRecord and VulnRecordVersion to ensure that the historical versions of the vulnerability are preserved and hence all the immutability of the historical scan reports. 
This will however increase storage space for the same vulnerability. How often do vulnerability definitions change?

proposals/new/common_vulnerability_schema_proposal.md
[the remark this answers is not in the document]
prahaladdarkin: Discussed with @danielpacak. Will retain the `PackageVersion` field.

proposals/new/common_vulnerability_schema_proposal.md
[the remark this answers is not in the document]
prahaladdarkin: @danielpacak  As discussed during our slack conversation, IMO for any scanner it is  the latest and greatest version of the CVE should be reflected across all reports to avoid ambiguity. It is better to be safe by raising flags to the user indicating that the vulnerability existed historically so that they can take corrective measures or validate any impact in retrospect. 
Additionally, I have now introduced a field for identifying the scanner within the CVE table. In this way each CVE is qualified by the CVE ID and the Scanner ID and hence would be unambiguous if multiple scanners are registered for a project or in the repository

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,238 @@
...
+	StatusCode       int       `orm:"column(status_code)"`
+	StatusRevision   int64     `orm:"column(status_rev)"`
+	StartTime        time.Time `orm:"column(start_time);auto_now_add;type(datetime)"`
+	EndTime          time.Time `orm:"column(end_time);type(datetime)"`
+}
+```
+
+As can be seen above the `Report` data structure no longer holds a JSON blob for the entire report. Each record is represented by the `VulnerabilityRecord` structure shown below
+
+### Create VulnerabilityReport data structure mapped to vulnerability_data table
```
danielpacak, line 88:
```suggestion
### Create VulnerabilityRecord data structure mapped to vulnerability_record table
```

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,238 @@
...
+// with the ID of the scanner that owns the CVE record definition.
+// The scanner ID would be the same as the RegistrationUUID field of Report.
+// Identified by the `cve_id` and `registration_uuid`.
+// Relates to the image using the `digest` and to the report using the `report UUID` field
+type VulnerabilityRecord struct {
+	ID               int64    `orm:"pk;auto;column(id)"`
+	CVEID            string   `orm:"unique;column(cve_id)"` 
+	ScannerID		 string   `orm:"unique;column(registration_uuid)"`   
+	Digest           string   `orm:"column(digest)"`
+	Report           string   `orm:"column(report_uuid)"`
```
danielpacak, line 102: Do we need the the `report_uuid` FK? Below you proposed a junction table, so if that's the case this would be redundant.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,238 @@
...
+// Since multiple scanners could be registered with the projects, each scanner
+// would have it's own definition for the same CVE ID. Hence a CVE ID is qualified along 
+// with the ID of the scanner that owns the CVE record definition.
+// The scanner ID would be the same as the RegistrationUUID field of Report.
+// Identified by the `cve_id` and `registration_uuid`.
+// Relates to the image using the `digest` and to the report using the `report UUID` field
+type VulnerabilityRecord struct {
+	ID               int64    `orm:"pk;auto;column(id)"`
+	CVEID            string   `orm:"unique;column(cve_id)"` 
+	ScannerID		 string   `orm:"unique;column(registration_uuid)"`
```
danielpacak, line 100: Maybe we should call it RegistrationUUID? It's the same value as we expect to be sent as the value of the Report.RegistrationUUID property, isn't it?

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+### Create a junction table to hold the vulnerability record per report
+```
+//ReportVulnerability is relation table required to optimize data storage for both the
+//vulnerability records and the scan report.
+//identified by composite key (ID, Report)
+//Since each scan report has a separate UUID, the composite key
+//would ensure that the immutability of the historical scan reports is guaranteed
+type ReportVulnerability struct {
```
danielpacak, line 123: Not sure if I understood this one. Is this a junction structure to associate a `Report` with zero to many `VulnerabilityRecords`? If yes, we may call it `ReportToVulnerabilityRecord` and prefix FK properties with `ReportID` and `VulnerabilityRecordID`?

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,238 @@
...
+```
+### CVSS Vector Information Modeling.
+
+In order to address the use cases of a holistic reporting of vulnerability data based on CVSS, a set of schema tables that would model the (CVSS3.x)[https://www.first.org/cvss/specification-document#Vector-String] and (CVSS2.x)[https://www.first.org/cvss/v2/guide] vectors is also proposed. The schema definitions are as follows
+
+```
+//CVSS3Vector is table that would store the CVSS 3.x related attack vectors
+//Identified uniquely by CVSS3VectorUUID
+//Also identified by composite key (CVEID, SourceUUID).
+type CVSS3Vector struct {
```
danielpacak, line 136: I'm not terribly sure if we should structure CVSS vectors as a first class relation in RDB model. At least not in the first iteration. Maybe it's good enough to start with properties / columns in the VulnerabilityRecord / vulnerability_record?

```go
type VulnerabilityRecord struct {
	// ...

	CVSS3Vector string // e.g. CVSS:3.0/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N
	CVSS2Vector string // e.g. AV:L/AC:M/Au:N/C:P/I:N/A:N
}
```

danielpacak (dismissed):
LGTM

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,167 @@
...
+
+Additional advantage would be that the Postgres query planner would be able to optimize the queries required to generate reports or perform other analytics and hence would result in faster response times.
+
+Literature on the web also suggests that in general normalizing the data in tables will reduce the data storage requirements since redundant data such as keys in the JSON document are now first class column names and would not occupy redundant storage 
+
+## Data retention and clean-up
+
+There are no changes proposed to the existing data retention time in Harbor. The clean-up logic would have to consider the normalized schema structure and clean up the data for an image using elementary SQL operations.
+
+Additionally, Garbage collection process may have to be modified to clean up vulnerability data for garbage collected images
```
steven-zou, line 160: Maybe not handled by GC, just clean the vuln data when deleting the specified images.

steven-zou (dismissed):
LGTM

@prahaladdarkin 

I think this is the right way to handle the vuln data. With this schema refactor, it's very friendly to do ad-hoc query that is a feature we're planning to support later.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,167 @@
...
+### Create VulnerabilityRecord data structure mapped to vulnerability_record table
+```
+// VulnerabilityRecord of an individual vulnerability. Identifies an  individual vulnerability item in the scan.
+// Since multiple scanners could be registered with the projects, each scanner
+// would have it's own definition for the same CVE ID. Hence a CVE ID is qualified along
+// with the ID of the scanner that owns the CVE record definition.
+// The scanner ID would be the same as the RegistrationUUID field of Report.
+// Identified by the `cve_id` and `registration_uuid`.
+// Relates to the image using the `digest` and to the report using the `report UUID` field
+type VulnerabilityRecord struct {
```
steven-zou, line 97: @danielpacak 

Is it a common way for kinds of scanners to return the concrete image layer info that contains the found vulnerabilities? If I'm not mistaken, Clair has such info. What about Trivy and Aqua CSP?


cc @prahaladdarkin

danielpacak: In general we should consider LayerID (and LayerDigest) optional properties because not every pluggable scanner might support it. For Trivy and Aqua CSP we do return LayerID.

prahaladdarkin:
@danielpacak  and @steven-zou  - one query.. as i was working on a PoC to convert the raw report to the normalized schema, I see that the current table `scan_report` and the new schema (`scan_report_v2`) in this proposal are almost identical except that the new proposed table does not have the `Report` column.
So wanted to understand the common practice followed - would it be OK to re-use the original `scan_report` table itself instead of creating a new table again. IMO, we can re-use the existing table. This way we can allow the following:
* Allow the existing functionality and schema to co-exist with the new ad-hoc query functionality that will be introduced until we switch over completely to the new functionality.
* Reduce the amount of code change - all through the various layers right up to the REST APIs, we would not need to change the name of the table
* Avoid complex queries and data management - copying the content from existing `scan_report` to the new `scan_report_v2`, deleting and cleanup.

Let me know your views

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,171 @@
...
+
+Using a first class RDBMS data to store the data would facilitate the use cases of ad-hoc querying and analytics.
+
+The two steps proposed are:
+ * Update the existing `scan_report` table to eliminate the `Report` column with the `json` data type.
+ * Create a new table `vulnerability_record` that will contain the normalized vulnerability schema for the images.
+
+### Proposed changes to `Report` data structure and hence scan_report table
+
+As can be seen above the `Report` data structure no longer holds a JSON blob for the entire report. Each record is represented by the `VulnerabilityRecord` structure shown below and would be stored in a relational schema structure
```
danielpacak, line 62: I think we should clarify that we still use the `scan_report` table and associate it with zero to many `vulnerability_records` through the `report_vulnerability_record` junction table. In the current description we are mixing reports with vulnerability records / properties.

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,171 @@
...
+	VulnRecordID int64  `orm:"column(vuln_record_id);"`
+}
+```
+
+### Example Ad-hoc queries
+The types of ad-hoc queries that can be made possible using the above schema are illustrated below
+
+### Number of software artifacts with a particular CVE-ID across all projects
+
+`select count(1) from vulnerability_record where CVEID = '<CVE_ID>'`
```
danielpacak, line 114:
```suggestion
```sql
select count(1) from vulnerability_record where cve_id = '<CVE_ID>'`
```

danielpacak: Maybe it's worth mentioning that this query will return CVEs from all registered scanners. Not the default system-wide scanner or the one set at the project level

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,171 @@
...
+
+### List packages and corresponding images:tag with particular CVE-ID across all projects
+
+```
+select  vr.package, vr.package_version from vulnerability_record vr inner join report_vulnerability_record rvr on vr.id = rvr.vuln_record_id
+                                          inner join scan_report sr on rvr.report_uuid = sr.uuid and sr.uuid in (select uuid from scan_report sr inner join
```
danielpacak, line 120: Can we just extend the outer WHERE predicate instead of using embedded SELECT statement?
I think this will be more readable. Also we can use sql syntax to markdown and fix indentation for this example.

danielpacak (changes requested):
:wave: @prahaladdarkin I left a few comments to tidy up this proposal and align it with the actual Go code.

proposals/new/common_vulnerability_schema_proposal.md
prahaladdarkin: Done

proposals/new/common_vulnerability_schema_proposal.md
prahaladdarkin: Done

proposals/new/common_vulnerability_schema_proposal.md
```diff
@@ -0,0 +1,171 @@
...
+## Technical Description
+An image consists of vulnerabilties or non-compliance of different types. They are largely categorized as below:
+* Operating System and software package related
+* Image structure related
+* Image file content related
+* License related
+
+A non-compliance or vulnerability is said to occur whenever there is discrepancy observed between the desired and actual results of a scan triggered for the above type of data.
+
+For e.g., a certain project within Harbor would require that all it's images have packages with the MIT license. The presence of a GPL packages within the container image would trigger a non-compliance.
```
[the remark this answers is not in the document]
prahaladdarkin: Done

prahaladdarkin (commented):
Done. Incorporated all review comments and resolved outstanding conversations.

danielpacak (approved):

steven-zou (approved):
The feature will be delivered in V2.2.

PR #150, opened 2020-11-10
Implementation of per robot account IP allowlist

joschi36:
I want to gather a little feedback regarding this one in the community.
For us, it would give more security as we can more granular control it.

Also, I don't know if the complexity to do a change like this in the codebase. Maybe someone can give me an assessment.


joschi36:
@steven-zou @ywk253100 do you have me some feedback?

joschi36:
I will close this as I've got no feedback and already got a different solution. I've created an ingress per project with different whitelist source ranges.

PR #157, opened 2021-03-15
add multi arch support

yuzp1996:


yanji09:
Would you add the Signed-off information for this commit to pass the DCO? @yuzp1996

PR #158, opened 2021-03-30
Add proposal for init config by env

stonezdj:
Add proposal for init Harbor configure by the environment variable.
EPIC https://github.com/goharbor/harbor/issues/14494


proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Implementation
+
+Update the common/config/manager.go 's [configFromEnv()](https://github.com/goharbor/harbor/blob/6624c136bac4705e06afe0667683f871ee12741f/src/common/config/manager.go#L118)
+
+When there is an INIT_CONFIG_JSON env provided, all user configure is marked to read-only and only can be initialized by this environment variable.
+
+Any one of the user's configure items are set by the environment variable, such as LDAP_URL is configured in the environment variable, then all user settings should be updated by the environment variable. The [metadatalist.go](https://github.com/goharbor/harbor/blob/4b033c266a6ff254f3b9a9cb59b7226499e0f42d/src/common/config/metadata/metadatalist.go#L64) defines the environment to set environment variable.
+
+When user configure items are initialized by the environment variable, they should be stored in the database table "properties" and if there is an existing config in the table properties, it checks if the auth_mode will be changed, if there is any user exist in the harbor_user table, then the auth_mode cannot be updated, there should be an initialization error in core container.
```
reasonerjt, line 55: I think generally there should be some checks before the configuration is persisted in the DB.

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,70 @@
...
+
+If the previous instance is installed with an INIT_CONFIG_JSON provide, this environment is lost after the core container restart, then it checks if the INIT_CONFIG_JSON provided, and failed the core container with the error message "Previous core container is initialized by INIT_CONFIG_JSON, it cannot start without the INIT_CONFIG_JSON.
```
reasonerjt, line 57: Is it necessary?  If the configuration is persisted to DB, next time if the JSON is not provided, the previous values, i.e. the configurations in the DB can be used?

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,70 @@
...
+
+If there are two core instances, which use the same database store, it might be the latest core container's INIT_CONFIG_JSON overwrite that of the previous core container.
```
reasonerjt, line 59: We should also emphasize this is an erroneous situation.  And this is not expected in normal use case.

reasonerjt:
After reading the proposal I feel the `init_config` may be better called `config_overwrite` ?  b/c it will always overwrite the configuration in DB when the instance is started.

We should also cover the case upgrade use case where the CONFIG_JSON is set for a env which already has configuration values in the DB.

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,70 @@
...
+Before Harbor 1.8, it is easier to configure external authentication by environment variables. After Harbor 1.8, it is could be configured by REST API to change the configuration, but it is not easy to issue the command and check the result in DevOps. and it is inconvenient when the Harbor instance is deployed in the Kubernetes environment. as discussed in issue [#8076](https://github.com/goharbor/harbor/issues/8076)
+
+
+## Proposal
+
+We can enhance the current configuration by adding some environment variables to initialize the Harbor configuration.
+
+INIT_CONFIG_JSON='' which contains the JSON format of the init configuration, for example:
+```
+INIT_CONFIG_JSON={"auth_mode":"ldap","ldap_base_dn":"dc=example,dc=com","ldap_base_url":"ldap.myexample.com"}
```
reasonerjt, line 20: +1 for the JSON.

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,70 @@
...
+All configure is not configurable anymore when there is a configure setting detected. make the user configure item partially editable is not in the scope of the current implementation.
+To change settings, users need to update the environment variable and restart Harbor.
+
+## Implementation
+
+Update the common/config/manager.go 's [configFromEnv()](https://github.com/goharbor/harbor/blob/6624c136bac4705e06afe0667683f871ee12741f/src/common/config/manager.go#L118)
+
+When there is an INIT_CONFIG_JSON env provided, all user configure is marked to read-only and only can be initialized by this environment variable.
+
+Any one of the user's configure items are set by the environment variable, such as LDAP_URL is configured in the environment variable, then all user settings should be updated by the environment variable. The [metadatalist.go](https://github.com/goharbor/harbor/blob/4b033c266a6ff254f3b9a9cb59b7226499e0f42d/src/common/config/metadata/metadatalist.go#L64) defines the environment to set environment variable.
```
reasonerjt, line 53: If you use `config_json` approach it's easier because you only need to check one env var.

proposals/new/init_config_by_env.md
reasonerjt, line 55: I think we need to think and clarify at which point the configuration values will be written to the DB.  They should be set in a relatively early stage in the bootstrap process of `harbor-core`.  Before it starts serving the API and probably also before the async jobs are scheduled.

proposals/new/init_config_by_env.md
stonezdj: Updated

proposals/new/init_config_by_env.md
stonezdj: agree

stonezdj: updated

proposals/new/init_config_by_env.md
stonezdj: Removed

proposals/new/init_config_by_env.md
stonezdj: Update

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,58 @@
...
+
+All user configure is not configurable anymore when there is a overwrite configure setting detected. make the user configure item partially editable is not in the scope of the current implementation.
+To change settings, users need to update the environment variable and restart Harbor.
+
+## Implementation
+
+Add an OverwriteConfig method to the config controller, and call this method in the main function, after database is ready and before any other API service or asnyc job start. When there is an CONFIG_OVERWRITE_JSON env provided, all user configure is marked to read-only and only can be initialized by this environment variable. It stops the core process for any error in the OverwriteConfig.
+
+When user configure items are overwrited by the environment variable, they should be stored in the database table "properties" and if there is an existing config in the table properties, it checks if the auth_mode will be changed, if there is any custom user exist in the harbor_user table, then the auth_mode cannot be updated, there should be an initialization error in core container. and all settings should have the same format with the PUT request for /api/v2.0/configurations.
```
reasonerjt, line 41: I think we need to make sure that the API and `OverwriteConfig` share the same code for validation.  So the CONFIG_OVERWRITE_JSON can  be seen as another input in addition to API request, if a request fails and the user puts the content into CONFIG_OVERWRITE_JSON it should fail with the same error.

I think we also need to clarify what happens if the validation for CONFIG_OVERWRITE_JSON fails, will `harbor-core` panic or the process for writing the configuration be skipped?

reasonerjt (approved):

proposals/new/init_config_by_env.md
stonezdj: The OverwriteConfig will verify the configure with the same logic as the API request before updating.
If the validation fails, the core process exit with the error message.

proposals/new/init_config_by_env.md
paulczar: I would be hesitant to store config details in a database, that will mean you're potentially storing admin level credentials in the database which would increase the attack surface area right?  I'm not a security expert, but it seems like a potential issue.

paulczar:
I think this is missing the underlying issue.  I feel strongly that System level config settings should be set at the system level ( config files, cli args, env variables) and app level settings should be set in the app ( api / harbor admin gui ).

paulczar:
People may also want the ability for this to be completeing configurable by config, for instance if there is a known set list of projects / users ...  especially for things like edge clusters, then setting base settings as well as a list of repositories in config would be desirable.

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,58 @@
+# Proposal:  `Overwrite configuration by environment variables`
+
+Author: `stonezdj`
+
+## Abstract
+
+It proposes an approach to Overwrite Harbor configuration by the environment variable.
```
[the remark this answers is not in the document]
stonezdj: Updated

proposals/new/init_config_by_env.md
stonezdj: These credentials are stored in the database with hash and salt, there is no change to the previous implementation.

stonezdj:
> I think this is missing the underlying issue. I feel strongly that System level config settings should be set at the system level ( config files, cli args, env variables) and app level settings should be set in the app ( api / harbor admin gui ).

The current proposal doesn't change the scope of configurations, the system and user configurations are still as they were.

proposals/new/init_config_by_env.md
```diff
@@ -0,0 +1,58 @@
...
+The configure panel in read-only mode is the following.
+
+![Configure Pannel](../images/init-config/configure.png)
+
+## Goal
+
+The overwrite config by env feature is only used for DevOps, when the overwrite configure env is provided, all user settings become read-only and can not be changed in configure pannel.
+All user configure items are stored in the Harbor database.
+
+## Non-Goals
```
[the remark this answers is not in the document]
stonezdj: Remove/add the env CONFIG_OVERWRITE_JSON can do the switch.

reasonerjt (approved):

slinstaedt:
Sorry for showing up, but not being invited to the party. 

Though this proposal will solve a quite some issues with infrastructure automation, the implementation is questionable: I have never seen an app, using a _single_ environment variable for the whole configuration. That is normally done using a config file, which are bind mounted into containers. 12factor apps [1] is also mentioning "env var**s**". Even infamous solutions like Spring's `SPRING_APPLICATION_JSON` never made it into real life usage for obvious reasons: Though it might be comfortable from the app provider side, to simple deserialise a single json string, app consumers (e.g. system administrators) have to take extra care of encoding and especially escaping everything correctly into this single string.

I would rather suggest using multiple env vars with special ones for credentials, each containing a single config value and/or using a container-local config file, configured via a single env var or cli parameter, like others players (traefik, drone, docker registry, keycloak, graylog). This will also provide better security, because env vars are known for being leaky.

[1] https://12factor.net/config

Vad1mo:
I fully agree with @slinstaedt, An environment variable with a JSON config block is uncommon. We will also complicate our and users' life when we they try to modify any of the values in the JSON via Helm Chart or docker-compose. 

Can we use the existing `harbor.yml` and the matching ENV VARs setup that alredy exists? This way we would have the same consistency and workflow as people already know

lindhe:
@stonezdj  Are you willing to discuss this proposal? I strongly feel this is a bad design decision that will hurt the project moving forward.

stonezdj:
> @stonezdj Are you willing to discuss this proposal? I strongly feel this is a bad design decision that will hurt the project moving forward.

It is an optional feature, if you don't like it, just ignore it.

lindhe:
Yeah, it's optional. But when a new feature is introduced, people will begin using it. And that makes migrating away from it a very painful process, once the issues with this implementation becomes clear. That's why it's a better idea to discuss it first and implement it later, once we have reached some sort of consensus on the design.

I assumed this was posted as a _proposal_ in the _community_ repository specifically so the _community_ (you maintainers as well as we users and others) are able to discuss the feature together and suggest changes _before_ the proposal is implemented. What's the point of this proposal otherwise?

If I have misunderstood the way of working, please explain to me how we as a community are supposed to interact with the project. Surely it's better to discuss proposals beforehand, so we can avoid changing implementation causing breaking changes?

stonezdj:
> Yeah, it's optional. But when a new feature is introduced, people will begin using it. And that makes migrating away from it a very painful process, once the issues with this implementation becomes clear. That's why it's a better idea to discuss it first and implement it later, once we have reached some sort of consensus on the design.
> 
> I assumed this was posted as a _proposal_ in the _community_ repository specifically so the _community_ (you maintainers as well as we users and others) are able to discuss the feature together and suggest changes _before_ the proposal is implemented. What's the point of this proposal otherwise?
> 
> If I have misunderstood the way of working, please explain to me how we as a community are supposed to interact with the project. Surely it's better to discuss proposals beforehand, so we can avoid changing implementation causing breaking changes?

First of all, I agree to the @slinstaedt 's comments, but the PR goharbor/harbor#14777 was merged in May 12 and this feature is already released in Harbor . The proposal should be consistent with the implementation. and we should close the current discussion.
Our intension is to add feature sort of MVP and enhance it step by step, I created https://github.com/goharbor/harbor/issues/15551  to follow up the issue.

wy65701436 (approved):
lgtm

steven-zou (approved):
lgtm

reasonerjt (approved):

wy65701436 (approved):
lgtm

PR #163, opened 2021-05-12
proposal bandwidth limit

bitsf:
for issue https://github.com/goharbor/harbor/issues/13194


proposals/new/bandwidth-limit.md
```diff
@@ -0,0 +1,135 @@
...
+- other level except job level would be much harder to implement as it need to schedule and collaborate between processes and instances
+
+## Compatibility
+
+the default behavior is no speed limiter, use default value `0` as unlimit.
+
+## Implementation
+
+### UI
+- user can specify speed on UI when edit replication rule
```
reasonerjt, line 37: potentially there may be 1:n mapping between policy and job so we need to clearly specify in the doc and on UI, that this number is applied to each job or the whole n concurrent jobs mapped to one policy.

steven-zou:
Using fixed speed limitation in policy rules seems a little inefficient but simple.

AllForNothing (approved):
Both UI and back-end  are finished  and implemented the target

ywk253100 (approved):

PR #164, opened 2021-05-23
Proposal for Unified Pluggable Scanner API

prahaladdarkin:
Scope of change:

Introduced the Unified Pluggable Scanner API proposal based on the discussions in the Harbor Scanner Workgroup.

Signed-off-by: prahaladd <prahaladd@vmware.com>


proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
+
+# Proposal: Unified API for Pluggable Image Scanners
```
danielpacak, line 2: Not sure if this title suggests the ultimate goal of this proposal.

IIRC we wanted to leverage API v1.1 which is already defined as a minimal set of API endpoints and leverages MIME types. IMO the main focus should be around new MIME types and new scanners / security tools to onboard

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+curl -XPOST http://scanner-adapter:8080/api/v2/scan -H 'Content-Type: application/vnd.scanner.adapter.scan.request+json; version=2.0' -d {
+         "registry": {
+           "url": "harbor-harbor-registry:5000",
+           "authorization": "Bearer: JWTTOKENGOESHERE"
+         },
+         "artifact": {
+           "repository": "library/mongo",
+           "digest": "sha256:917f5b7f4bef1b35ee90f03033f33a81002511c1e0767fd44276d4bd9cd2fa8e"
+         }
+         "scanTypes": {
```
danielpacak, line 88: The scanType property is a subset of produced MIME types advertised by the /metadata endpoint so we can think about different / consistent naming here. More importantly, I'd use an array of objects instead of a map with MIME types as a key to enable further extensibility:

```json
{
  "scanTypes": [
    {
      "mimeType": "application/vnd.security.vulnerability.report; version=1.1",
      "params": {}
    },
    {
      "mimeType": "application/vnd.security.bom.report; version=1.0",
      "params": {
        "allowGNU": false,
        "format": "spdx"
      },
      "propertyAddedInTheFuture": {}
    }
  ]
}
```

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+       }
+```
+In the above request the `scanTypes` key refers to an object that contains key-value pairs with the MIME type as a key.
+ - Each type of scan is identified by it's MIME type - in the above request we are requesting for vulnerability scan and BoM scan from a scanner.
+ - Each type of scan can be further controlled or additional params passes in using the key value pairs. For e.g. in the above sample request, the BoM scanner can be controlled using the keys allowGNU and the format of representing the BoM information (in this case spdx).
+ - Scanner can be requested to perform only a specific type of scan by including only the specific corresponding MIME type in the request along with any control parameters.
+
+The response from the scanner are packaged in an uber-container structure again keyed by the MIME type. The object against each MIME type would contain the following
+   - Internal scan tracking identifier for the composite scan.
+   - The scan id and status for each MIME type. The status would be either one of started, completed, running, aborted, failed
```
danielpacak, line 101: I don't think we need this complexity. It's non backward compatible and requires designing and implementing a workflow to rescan for a particular MIME types. I believe that most of the time a scanner adapter will focus on one MIME types. For scanners that support multiple MIME types it's probably acceptable to retry all scan types.

Beyond that, scanner adapters will have to maintain composite identifiers to implement state machine for each scan type.

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+* Validate the MIME types present in the request.
+  * Check the MIME type is in the list of known MIME types
+  * Check that there is a scanner available for specified MIME type
+* For each MIME type in the scan request payload, in a non-blocking manner:
+  * Send the scan request to the registered scanners for the MIME type
+  * Collect the scan result (or error) for the MIME type.
+* Create an uber-container data structure which will contain the scan results keyed by the corresponding MIME types and send the result to client.
+
+### Scan API response handling
+The scanning process would be executing multiple types of scans asynchronously and hence a sophisticated response handling mechanism would required.
```
danielpacak, line 135: In the example above there's one scan request with multiple scanTypes indexed by mime type. On the other hand we'd be tracking multiple scan identifiers embedded in the scan response. Why not sending multiple scan requests and tracking one by one? Or if we stick to one request with multiple scanTypes we could track all mime types as a whole

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+
+For a given job, the multiplexer component would maintain a mapping between the composite scan Id and the granular status of each of the component scans  as and when received from the scanner.
```
danielpacak, line 137: For some scanners keeping such granularly and scan id structure might be overly complex.

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+### Error Handling
+
+The above API invocation pattern requires an enhanced error handling mechanism since multiple scan jobs of different types would be executing and some of them could potentially error out. There are three important aspects to be considered in error handling
+  -  how does the scanner report an error to Harbor core?
+  -  how is the error reported to the caller?
+  -  what is the impact of an error on the overall scan job?
+
+A clean, simple and efficient mechanism to handle errors would be as follows
```
danielpacak, line 151: We could return an array of errors. Each error might have a mimeType or scanType property to indicate why it's failing

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+ - If a scanner reports an error when scanning for a particular MIME type, then the uber-container response to the client will contain the serialized error against the MIME type.
+ - Each scan is mutually exclusive of the other and the failure of one does not affect the other. The response of a composite scan job submission would contain the error details for each MIME type and it is the responsibility of the client to handle the error appropriately.
+
+
+## Advantages of proposed design
+1. Prevents combinatorial explosion of REST endpoints for each type of scan (and/combinations) to be supported.
+2. Adopts a "what you send is what you receive" model.
+3. Allows clients to control what data they want from the scanner adapter - when the same scanner supports multiple types of scan functionalities
+4. Provides a mechanism to fine tune scanning operation.
+5. Incremental changes on the existing API model exposed by Harbor.
```
danielpacak, line 163: Complex structure of scan id per mime type would required lots of changes in Harbor core / jobservice

danielpacak (commented):
👋 @prahaladdarkin This is great start. I left a few comments / asks for clarification. Overall I'd focus more on onboarding new tools and keep API structures as close to v1.1 as possible without breaking compatibility.

Adding more examples of a hypothetical adapter that supports both vulnerability scan and licence information, including happy path and partial errors would be very helpful to understand the end to end integration.

prahaladdarkin (commented):
Responded to some review comments with some clarifications. We can discuss in detail in the meeting and then I will make the required changes.

proposals/pluggable-image-scanning-api-v2.md
```diff
@@ -0,0 +1,177 @@
...
+The approach for designing a unified API surface is detailed below. 
+
+### Supporting multiple scan types
+
+There would be one MIME type for the result data of each scan type. A few MIME types are proposed below:
+
+| Mime Type | Type | Version | Status | 
+| --- | --- | --- | --- |
+| application/vnd.security.vulnerability.report; version=1.1 | OS/Package vulnerability data | 1.1 | Released in Harbor 2.3.0 |
+| application/vnd.security.bom.report; version=1.0 | Bill of Materials (BoM) | 1.0 | Proposed |
```
zhill, line 67: There are several near-standard formats that would make sense for SBoM instead of a new proprietary one: SPDX and CycloneDX are the two with the most traction. Those will help with interoperability, and may include enough information for the FS verification type as well (SPDX in particular supports files and digests as long as you are ok with very large manifests.

PR #166, opened 2021-05-24
Backup Hosting Harbor Images in TCR

kofj:
WIP. Update later.

Signed-off-by: fanjiankong <fanjiankong@tencent.com>


Vad1mo:
We close for now, this proposal for now.  Due to the current lack of interest from the community. However, this decision does not prevent the possibility of revisiting your proposal in the future, should the situation change.

PR #169, opened 2021-06-09
Add draft proposal for project auto sync

yunkunrao:
Signed-off-by: yunkunrao <yunkunrao@gmail.com>


OrlinVasilev:
@reasonerjt @steven-zou as technical leads can you please review such proposals!

That one was submitted 1 year ago and not a single comment :( 

Thanks you!

zyyw:
@yunkunrao I'm wondering why you would like this project-auto-sync instead of using replication to achieve the sync purpose.

PR #172, opened 2021-07-27
add proposal cosign integration

wy65701436:
The proposal is for cosign and harbor integration story

Signed-off-by: Wang Yan <wangyan@vmware.com>


dlorenc:
This is awesome and I would love to help out anyway possible! (I'm a maintainer on cosign)

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+ artifact_id int,
+ /*
+  the subject artifact id of the accessory.
+ */
+ subject_artifact_id int,
+ /*
+  the type of the accessory, like signature.cosign.
+ */
+ type varchar(1024),
+ digest varchar(1024),
```
reasonerjt, line 83: is this column needed?  it's a dup of artifact.digest?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+         "urls":null
+      }
+   ],
+   "accessories":[
+      {
+         "artifact_id":1,
+         "subject_artifact_id":81,
+         "digest":"sha256:94788818ad901025c81f8696f3ee61619526b963b7dc36435ac284f4497aa7ca",
+         "type":"signature.cosign",
+         "icon":"sha256:0048162a053eef4d4ce3fe7518615bef084403614f8bca43b40ae2e762e11e06",
```
reasonerjt, line 215: I may need to refresh my memory about this but do all the icons stored as blobs now, even the preset ones like container images?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+### Request Artifact Accessory
+
+**List artifact**
+
+1. You can use the `signature.cosign` in the accessories to determine whether an object is signed and to view information about the signature, and you can pull the signature with digest in the registry store, 
+2. You cannot list the accessories directly via list artifact API.
+
+```yaml
+
+GET /api/v2.0/projects/library/repositories/hello-world/artifacts
```
reasonerjt, line 167: Will need to add a querystring for accessories?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+X-Request-Id: 8d70f12a-9f73-4fc9-9e81-3bed24926254
+Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
+X-Frame-Options: DENY
+Content-Security-Policy: frame-ancestors 'none'
+```
+
+**Delete artifact signature(specific)**
+
+```rest
+DELETE /api/v2.0/projects/library/repositories/hello-world/artifacts/sha256:1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792/accessory?type=signature.cosign&digtest=sha256:94788818ad901025c81f8696f3ee61619526b963b7dc36435ac284f4497aa7cb
```
reasonerjt, line 298: what if the `sha256:94788818ad901025c81f8696f3ee61619526b963b7dc36435ac284f4497aa7cb` is not a accessory of `sha256:1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792`?

Do you wanna support both digest and artifact_id when managing an accessory?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+### Managing Signed Object
+
+Use this information to know about how you can work with signed objects and how signed object affect system behaviors. Even if you do not sign objects youself, you need to understand how to work with them, and how these signed objects affect routine system administrative tasks.
+    
+#### Replication
+
+- You will have an option indicate that whether you want to replicate the signature between Harbor instances.
+    1. Replication the artifact to the target Harbor.
+    2. Replication the signature to the target Harbor, then the target Harbor will build the reference.
```
reasonerjt, line 381: Since you introduced the `accessory` abstraction layer you should describe the replication scenario in `accessory` terms.

For simplicity if possible IMO all accessories should be replicated considering artifact+accessories is one entity.

What if the target Harbor does not set to `cosign` mode?  I think in that case the signature should not be replicated?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+![Signature_replication](../images/cosign/signature_replication.png)
+
+#### Tag Retention
```
reasonerjt, line 385: I don't think this is needed, this is an enhancement to tag retention policy that is out of the scope of cosign integration.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+#### Robot Account
+
+- You can use the robot account to pull the signature if it has pull permission.
+
+### Behaviors Difference (Notary & Cosign)
+
+1. Signed images:
+    - Notary: If user sign an image with notary and restart Harbor without notary enabled, the image will be acted as un-signed.
+    - Cosign: If user sign an image with cosign, unless user removes the signature from the artifact, the image will be acted as signed.
```
reasonerjt, line 428: So we wanna support notary and cosign at the same time?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+X-Request-Id: 8d70f12a-9f73-4fc9-9e81-3bed24926254
+Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
+X-Frame-Options: DENY
+Content-Security-Policy: frame-ancestors 'none'
+```
+
+### Signature checker Interface
+
+```go
+type PublicKey crypto.PublicKey
```
reasonerjt, line 316: What is the `PublicKey` related stuff?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+![Build Reference_Copy](../images/cosign/build_reference_copy.png)
+
+- copy artifact to the target project
+- copy signature to the target project 
+- build the reference in the database
+    
+
+### Request Artifact Accessory
```
reasonerjt, line 158: I think we need a section for API change, note that we also need to determine what to do with the existing APIs:
Do we list accessories in GET /artifacts API?
What about DELETE ...

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+}
+```
+
+User can specify anything for the tag attribute, even the tag does not exist. 
+
+Do we need to support this scenario?
+
+4. **Index Signing**
+
+Cosign can sign a multi-platform `Index`,  but the signature is attached to the index digest. Should we see all the references are signed?
```
reasonerjt, line 475: It's hard to handle that when you check the signature in a pull command?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+### DB scheme
+
+```yaml
+
+CREATE TABLE artifact_accessory (
+ id SERIAL PRIMARY KEY NOT NULL,
+ /* 
+    the artifact id of the accessory itself.
+ */
+ artifact_id int,
```
reasonerjt, line 74: I feel seeing accessory as an artifact make things more complicated in the API to handle artifacts you need to add a lot of `if` statements b/c accessories are different from regular artifacts.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+    - Cosign: Cosign attached the signature to the artifact digest(by default).
+
+3. Artifact Deletion:
+    - Notary: You cannot remove an artifact without first deleting its signature.
+    - Cosign: You can remove an artifact with its signature as you like.
+    
+### To be discussed
+
+1. Cosign attaches signature to digest by default, should we apply the signature to all tags of an artifact?
+2. Cosign can specify the signature repository, do we support it?
```
reasonerjt, line 441: Need more context about `signature repository`

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+3. Can the signature be added/removed tag or label like other artifact?
```
reasonerjt, line 442: It's something to think about for all accessories, see my comment in the table schema for accessory.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+  },
+  "Optional": {
+    "tag": "nothing"
+  }
+}
+```
+
+User can specify anything for the tag attribute, even the tag does not exist. 
+
+Do we need to support this scenario?
```
reasonerjt, line 471: We need to extract the blob to support it right?  If the answer is yes, I suggest not.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+4. **Index Signing**
+
+Cosign can sign a multi-platform `Index`,  but the signature is attached to the index digest. Should we see all the references are signed?
+
+5. **Multiple Signatures**
+
+- Immutable Repository
+
+Cosign's model of "appending" signatures won't work for multiple signatures.
```
reasonerjt, line 481: Do you mean if a repository is immutable in Harbor you can't push any signatures?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+PRECONDITION: Failed to process request due to 'hello-world:sha256-1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792.sig' configured as immutable.
+```
+
+Issue: [Design: Cosign support for registries with immutable tags](https://github.com/sigstore/cosign/issues/313)
+
+- Garbage Collection
+
+As the multiple signatures share one tag, the previously pushed becomes the untagged artifact. If user run GC with untagged checked, all of these untagged signatures will be removed.
+
+```json
```
reasonerjt, line 496: need more context about this JSON.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+      "annotations": {
+        "dev.cosignproject.cosign/signature": "MEUCICVyBbxVgsbJUiZJRbf/NaqSHjRd5V9UDzpr1oikXcyyAiEA4uZHYANWHkLSRIDd9+6ynV0tQ21n82+ArcvOUsw13EI="
+      }
+    }
+  ]
+}
+
+```
+
+Solution: For GC, the untagged artifact means that one artifact without any tag and are not an accessory.
```
reasonerjt, line 526: Then you'll need to make sure the accessory table is in sync, which can be messy to handle.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+type PublicKeyProvider interface {
+    /*
+      Read the public key from given artifact.
+    */
+    PublicKey(ctx context.Context, artifact Artifact) (PublicKey, error)
+  
+    /*
+      Read all the public key from system.
+    */
+    PublicKeyes(ctx context.Context)  ([]PublicKey, error)
```
ncdc, line 328:
```suggestion
    PublicKeys(ctx context.Context)  ([]PublicKey, error)
```

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+type Signature []byte
+
+type PublicKeyProvider interface {
+    /*
+      Read the public key from given artifact.
+    */
+    PublicKey(ctx context.Context, artifact Artifact) (PublicKey, error)
+  
+    /*
+      Read all the public key from system.
```
ncdc, line 326: Do you want this at the system level, or per repository, or maybe both?

proposals/new/cosign-integration.md
ncdc: I think this makes sense.

proposals/new/cosign-integration.md
ncdc: I believe this is a client-side option only and wouldn't affect the server-side aspects of Harbor?

See https://github.com/sigstore/cosign#specifying-registry for info on what this does.

proposals/new/cosign-integration.md
ncdc: I think it makes sense to check that any content you're using as a consumer is signed - i.e. the manifest list (index) and any arch-specific content too.

proposals/new/cosign-integration.md
ncdc: You'd be able to push an initial signature manifest, but not any additional signatures.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+The Cosign signatures are stored in an OCI registry next to the container image, and can be located via a simple name scheme. Thus, Harbor can build up the relationship between the artifact and its Cosign signature in the database, and the Harbor function can use this relationship to determine behavior.
+
+## Goal
+
+1.  Support cosign sign artifact.
+2.  All the existing functionality can be applied to the signed artifact with cosign.
+
+## Non Goal
+
+1.  Do not manage Public keys in Harbor side.
```
ncdc, line 29: Is this a future goal?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+* User Stories
+
+1.  As a project admin & user, I can use cosign to sign an artifact(single or multiple times).
+2.  As a project admin & user, I can delete a cosign signature via Harbor UI/API.
+3.  As a project admin & user, I can delete/GC an artifact as well as its cosign signature.
+4.  As a project admin & user, I cannot GC an artifact's cosign signature individually.
+5.  As a system admin, I can copy an artifact and its signature to another repository.
+6.  As a system & project admin, I can reserve an artifact and its signature via retention policy.
+7.  As a system & project admin, I can set the content trust policy to block the un-signed artifact pulling.
```
ncdc, line 51: Without storing public keys, Harbor won't be able to verify the signatures. Would you allow a user to pull an image with a signature when Harbor can't tell if the signature is valid or not?

ncdc (commented):
Thanks for writing this up!

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+    - Notary: Notary attaches the signature to the image tag.
+    - Cosign: Cosign attached the signature to the artifact digest(by default).
+
+3. Artifact Deletion:
+    - Notary: You cannot remove an artifact without first deleting its signature.
+    - Cosign: You can remove an artifact with its signature as you like.
+    
+### To be discussed
+
+1. Cosign attaches signature to digest by default, should we apply the signature to all tags of an artifact?
```
ncdc, line 440: I wouldn't do this. The signature was created for an image with a specific digest (potentially associated with a specific tag at a point in time).

reasonerjt: He means the case where different tags associate with one digest.

ncdc: Oh, thanks for the clarification. In that case, I would say yes, display the signature for all tags pointing to the same digest.

proposals/new/cosign-integration.md
reasonerjt: Valid or not depends on whether a client chooses to trust the public key or not.

This policy may only check if the signature exists, with the cosign model whether the client should trust it or not is beyond the scope of Harbor.

proposals/new/cosign-integration.md
reasonerjt: Not likely unless tool chain supports push a public key as OCI artifacts.

proposals/new/cosign-integration.md
ncdc: I think it could go either way. Harbor could block if no signature exists and allow if any signature exists.

It could also go further and block if it can't verify the signature. It mostly depends on if we want to require public keys to be uploaded to Harbor so Harbor can verify.

For a first step (and possibly only), we can go with the first approach (Harbor doesn't verify signatures because it doesn't know about public keys).

proposals/new/cosign-integration.md
ncdc: What about a Harbor API for it?

ncdc: Or pubkeys as artifacts could work too

proposals/new/cosign-integration.md
steven-zou: +1

proposals/new/cosign-integration.md
steven-zou: What's the use case of this method?

proposals/new/cosign-integration.md
steven-zou: If the root artifact is successfully replicated and the signature artifact replication is failed, is that counted as a success or a failure?

proposals/new/cosign-integration.md
steven-zou: For integration with Harbor, living in the same repository seems much better and simpler.

proposals/new/cosign-integration.md
steven-zou: IMO, as it's an accessory of the root artifacts, it seems there are no use cases to require treating accessory artifacts as general ones。 They're just additional things of the root artifacts.

proposals/new/cosign-integration.md
steven-zou: -1.

Conflict with the `To-be-discussed item 1?

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+         "annotations":{
+            "dev.cosignproject.cosign/signature":"MEUCIQD/imXjZJlcV82eXu9y9FJGgbDwVPw7AaGFzqva8G+CgwIgYc4CRvEjwoAwkzGoX+aZxQWCASpv5G+EAWDKOJRLbTQ="
+         }
+      }
+   ]
+}
+
+```
+
+![Build Reference](../images/cosign/build_reference.png)
```
steven-zou, line 122: Leverage the current metadata extracting processor framework?

steven-zou (commented):
It's cool.

proposals/new/cosign-integration.md
```diff
@@ -0,0 +1,536 @@
...
+
+1.  Support cosign sign artifact.
+2.  All the existing functionality can be applied to the signed artifact with cosign.
+
+## Non Goal
+
+1.  Do not manage Public keys in Harbor side.
+2.  Do not support multiple signers in a deployment. 
+3.  Do not support cosign clean/delete(as the delete operation leverages the tag deletion API which is not a required implementation defined by OCI spec, and Harbor does not allow it). 
+4.  Do not support cosign copy(cosign pushes the signature firstly to the destination, and then the artifact, Harbor cannot build the reference.)
```
renmaosheng: the description is already in the non-goal section, we don't need to add the " Do not" in the sub clause.

proposals/new/cosign-integration.md
wy65701436: we can assume that the transation of repliation should contain artifact and signature, and it's failure and the repliated artifact should be removed from the target.

proposals/new/cosign-integration.md
wy65701436: for the cosign scenario, we can add a default icon in Harbor.

proposals/new/cosign-integration.md
wy65701436: Just for further design.

proposals/new/cosign-integration.md
wy65701436: to support it, it will involve the permission things(AZ).

proposals/new/cosign-integration.md
wy65701436: All the signatures shares one tag name, if the immutable enabled, the following push will be rejected.

proposals/new/cosign-integration.md
wy65701436: it's a draft design and for the phase 2. But, I suppose it's at the system level.

rszumlakowski (commented):
Does there need to be a user story to confirm that end users will be able to see images with Cosign signatures in the Harbor User Interface?

Something like:

> As a project admin & user, I can see which tags for an artifact have valid cosign signatures

tianon:
The biggest concern I have with this proposal right now is that the "accessory" that's defined here is very similar to the references work that's happening in the OCI right now (for example https://github.com/opencontainers/tob/issues/96, which has a lot of links to other discussions too).

I'd hate to see Harbor implement something that's very close to something the OCI eventually ratifies, but specific to `cosign` and not quite matching the OCI implementation (creating the need for a migration process for both Harbor and the users of it, assuming the thing that gets ratified is compatible enough to be something we could migrate users to directly).

dlorenc:
I don't really have a preference either way on waiting, but I don't think it's safe to assume the OCI is actually close to ratifying anything. That link is to a draft proposal that's been open since April and still hasn't been submitted as a real proposal for a vote.

After that vote, the WG would then need to actually start, design and prototype different options. Then the spec changes would need to get written, debated and approved. Then enough registries would need to implement those changes for it to be safe for clients to rely on them. 

I'm (pessimistically) guessing it's going to be mid to late 2022 at the earliest before there's enough widespread adoption of the new changes for clients to rely on them.

The cosign spec is defined here, with a goal of working with existing registries; https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md and is already implemented by a few other clients so it's not really cosign specific.

tianon:
Thanks @dlorenc for providing your thoughts -- I guess that changes the question a little bit for me then. :smile:

Why does adding "cosign support" to Harbor require creating this special "accessory" type?  Is the goal of doing so to make it more usable/useful to end users, or is there a deeper reason?

dlorenc:
> Why does adding "cosign support" to Harbor require creating this special "accessory" type? Is the goal of doing so to make it more usable/useful to end users,

My understanding is that this is for extra features, like UI integration and policy support on top of the signatures. Basic sign/verify/upload/download operations all work with harbor out of the box.

wy65701436:
> Thanks @dlorenc for providing your thoughts -- I guess that changes the question a little bit for me then. 😄
> 
> Why does adding "cosign support" to Harbor require creating this special "accessory" type? Is the goal of doing so to make it more usable/useful to end users, or is there a deeper reason?

To add the `accessory` is just to let Habor knows the `relationship` of artifact and its signature. Otherwise, from the perspective of Harbor, the cosign signature is just another irrelevant artifact. 

Through the `relationship`, Harbor can treat the artifact and its signature as a whloe, and then determine the behavior of GC/Remove/Replication/....

SteveLasker:
JFYI, the [ORAS Artifacts Draft Specification](https://github.com/oras-project/artifacts-spec/) has been [released](https://github.com/oras-project/artifacts-spec/releases/tag/1.0.0-draft.1). It generalizes the way references can be pushed, discovered and pulled, enabling a standard across all distribution-spec based registries:

chlins (approved):
lgtm

stonezdj (approved):

proposals/new/accessory.md
```diff
@@ -0,0 +1,247 @@
+Proposal: Add abstract definition accessory
+
+Author: Yan Wang
+
+## Abstract
+
+Beside to store an artifact into registry, there are requirements on storing secure chain artifacts, like Software Bill of Materials (SBoM) and signatures to provide rich  rich functionality of an artifact.
+
+The proposal is to introduce an abstract definition -- accessory -- to enable Harbor to extend the capabilities for storing artifact with additional stuff.
```
Vad1mo, line 9: @tianon you as a native speaker, do you think the term **[accessory](https://www.merriam-webster.com/dictionary/accessory)** is correct here or would you use a better suited term?

I would like to critically question the "term", as it will stick on forever and generation of Harbor developers and users would need lookup the manual and the code just to understand what it means. 

IMHO an artifact **artifact attachment** or **attachment** would be a better suited term because: 

1. It is a common and well know term for an object that can be (optionally) attached to another object. It is known to many user and developers from email, documents and files. 
2. A good word that denotes the common use correctly will save thousand hours of time for many many users and developers.

Vad1mo (changes requested):

steven-zou (approved):
LGTM

proposals/new/accessory.md
wy65701436: thanks @Vad1mo and @tianon, firstly, we'd like to extend the `accessory` into more cases, like SBoM, Nydus(converted images).

I know the OCI or Oras community uses the `attachement` to describe the linkage of subject manifest and its things. IMO, in Harbor, I'd like to use different term to descirbe, we can say, the same thing to avoid confusion. A typical example is `label`,  in Harbor we have this term to help user to filter artifact. However, in docker community, labels are an key-value pair.

And since we've reached the RC stage and all the source code has been merged, we have to draw an conclusion soon.

PR #173, opened 2021-08-15
feat: stop scan job proposal

zyyw:
Signed-off-by: Shengwen Yu <yshengwen@vmware.com>


tianon:
This isn't new to this proposal, but starting a new scan job immediately removes/invalidates the results of any previous successful scan job.  I think this is relevant here because this new functionality highlights that by adding a new "Scan Job Cancelled" state that gets displayed -- my immediate thought seeing that in today's demo was wondering how we could get back that previous scan result.

Could we maybe defer actually removing the previous scan result until the new scan is finished so that if the new scan is cancelled the old report can still show?

It would definitely be useful to be able to see more historical data on scans (which results are new, etc), but that's a totally separate feature request (and I'm not asking for that here).

wy65701436 (approved):
lgtm

heww (approved):

reasonerjt (approved):

PR #181, opened 2021-11-24
Proposal: Framework for storing and managing  system generated non-artifact data within the registry

prahaladdarkin:
Signed-off-by: prahaladd <prahaladd@vmware.com>


proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+2. Storing the generated data artifact within the database table as a BLOB. The core service then reads the BLOB from the table - This approach is better than the approach 1. The cost of computation time and storage involved in storing short-lived data into the database would be significant. Additionally, reading large blobs of data from database fields in a streaming fashion is not supported. This implies that the entire blob data bytes should be buffered in memory in order to stream the data to the client. This would result in significant memory usage and could even crash the service containers/pods if the data volume is very large
+3. Store generated data artifact on a data volume mounted within the service container - This approach exposes the filesystem to the service. This will allow consumers to store data files using the standard filesystem APIs. However, within Kubernetes environments, Harbor mounts only Persistent volumes and these persistent volumes cannot be re-mounted in different services. Hence this option cannot be adopted.
```
Vad1mo, line 18: Joblogs are already stored on a file system by default, (it can however be switched to DB)

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
+## Abstract
+
+This proposal introduces a framework that exposes the registry as a backend store for system aritfacts created by various Harbor subsystems.
+
+## Use cases
+* Sharing of ephemeral data items between Harbor services, where-in the data is too large to be streamed over HTTP(s) calls.
```
Vad1mo, line 6: Which services are these?
Also if the data is to big for HTTP how should the end-user download the data if not over HTTP?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+
+For the above use case, the following are the candidates for :
+
+1. A stateful session-based REST API exposed between the job service and Harbor core that would be used to stream the data file. - This approach requires implementation of complex state management logic within the job service and core for *every functionlity* that has requirement for sharing ephemeral data.
+2. Storing the generated data artifact within the database table as a BLOB. The core service then reads the BLOB from the table - This approach is better than the approach 1. The cost of computation time and storage involved in storing short-lived data into the database would be significant. Additionally, reading large blobs of data from database fields in a streaming fashion is not supported. This implies that the entire blob data bytes should be buffered in memory in order to stream the data to the client. This would result in significant memory usage and could even crash the service containers/pods if the data volume is very large
```
Vad1mo, line 17: PG supports streaming with Large Object, 

https://www.postgresql.org/docs/current/lo-implementation.html
https://www.cybertec-postgresql.com/en/binary-data-performance-in-postgresql/

However there are since ages downsizes with storing  binary filey in a DB

Vad1mo (commented):
Could you give an example of the expected data sizes? 100MB or 100GB?
If the focus here is on text and CSV files with CVEs then those files can be compresses very efficiently with 90 to 99% compression rate.

see https://www.adaltas.com/en/2021/03/22/performance-comparison-of-file-formats/#results


Such an implementation makes sense in my eyes after exploiting all the other possibilities.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+* Does not contain an OCI compatible manifest for describing the artifact data layers.
+
+
+### Tracking System Data Aritfact Creation
+
+The following structure would be used to track the creation of system data artifacts within Harbor
+```go
+// SystemArtifact represents a tracking record for each system artifact that has been
+// created within registry using the system artifact manager
+type SystemArtifact struct {
```
wy65701436, line 37: since the SystemArtifact is stored as a blob in filesystem, will it be recored in table blob? If not, it will not impact the system store usage that published to harbor admin.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+3. Store generated data artifact on a data volume mounted within the service container - This approach exposes the filesystem to the service. This will allow consumers to store data files using the standard filesystem APIs. However, within Kubernetes environments, Harbor mounts only Persistent volumes and these persistent volumes cannot be re-mounted in different services. Hence this option cannot be adopted.
+
+Due to the  shorcomings of the above approaches, we explore the approach of using the registry as an intermediate store for system-generated data that needs to be shared across Harbor services. The next sections provide further details on this approach.
+
+### System Data Artifacts
+
+The proposal introduces the concept of a system data artifact. A system data artifact:
+* Is an arbitrary blob of data in text or binary.
+* Is created by a Harbor service to be consumed by other Harbor services.
+* Has a short lifespan - typically in minutes/hours but not in days.
```
wy65701436, line 27: why it's desgined by short lifespan? For the vul report case, it mean that user can only download the report in hours?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+// By using this framework, harbor components can create artifacts for
+// cross component data sharing. The framework abstracts out the bookkeeping 
+// logic involved in managing and tracking system artifacts.
+type SystemArtifactManager interface {
+
+	// Create a system artifact described by artifact record.
+	// The reader would be used to read from the underlying data artifact.
+	// Returns a system artifact tracking record id or any errors encountered in the data artifact upload process.
+	// Invoking this API would result in a repository being created with the specified name and digest within the registry.
+	Create(ctx context.Context, artifactRecord *SystemArtifact, reader io.Reader) (int64, error)
```
wy65701436, line 85: the imple is to call v2 put or patch blob?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+The below interface defines an API that is used by Harbor subsystems to create system data artifacts
+
+```go
+// SystemArtifactManager provides a low-level interface for harbor services
+// to create registry artifacts containing arbitrary data but which
+// are not standard OCI artifacts.
+// By using this framework, harbor components can create artifacts for
+// cross component data sharing. The framework abstracts out the bookkeeping 
+// logic involved in managing and tracking system artifacts.
+type SystemArtifactManager interface {
```
wy65701436, line 79: maybe we need exist method as well, which can be corresponded to v2 head blob.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+
+	// List all the system artifact records that match the criteria specified
+	// within the query.
+	List(ctx context.Context, query *q.Query) ([]*SystemArtifact, error)
+}
+```
+
+### Artifact Cleanup
+
+Every harbor component is responsible for clean-up of any system artifact data that it generates. In order to prevent abuse of the Harbor registry due to an erronous logic in creation or deletion of system data artifacts, there would be an artifact clean-up job that runs periodically every 24 hours and deletes all system data artifacts older than 24 hours.
```
wy65701436, line 99: Generally, how does the system artifact  cannot be cleaned by harbor garbage collection?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+The below data flowchart depicts the cleanup logic that would be executed periodically
+
+![System artifact cleanup logic](../images/system-artifact/system-artifact-cleanup.jpg)
+
+### Conformance to OCI Artifact Standard
+System data artifacts do not have corresponding Manifests and hence  *do not* conform to the OCI Artifact specification.
+
+### UI visibility of system artifacts
+
+System artrifacts are created by internal harbor components and hence are not visible within the Harbor UI to both the System Admin or the Normal user.
```
wy65701436, line 110: if there is no API exposed, it seems that these part of data are out of control for Harbor admin.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+	return "system_artifact"
+}
+
+func (sa *SystemArtifact) TableUnique() [][]string {
+	return [][]string{{"repository_name", "digest"}}
+}
+
+```
+
+### System Data Artifact CRUD API
```
wy65701436, line 68: for the vul report scenario, how does the System Aritfact is associated with an artifact? will it be appened into the list/get artifact API, if no, how does UI know it and how does user download the vul report?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+	// Also deletes the tracking record from the underlying table.
+	Delete(ctx context.Context, string, digest string) error
+
+	// List all the system artifact records that match the criteria specified
+	// within the query.
+	List(ctx context.Context, query *q.Query) ([]*SystemArtifact, error)
+}
+```
+
+### Artifact Cleanup
```
wy65701436, line 97: In the multiple instances deployment, since there is one cleanup per core, can it be concurrent executed? No read&write conflict？

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
+## Abstract
+
+This proposal introduces a framework that exposes the registry as a backend store for system aritfacts created by various Harbor subsystems.
+
+## Use cases
+* Sharing of ephemeral data items between Harbor services, where-in the data is too large to be streamed over HTTP(s) calls.
+*  Sharing of ephemeral data between Harbor services, where-in the data is too large to be stored within the database or to be read from the database.
```
steven-zou, line 7: Not just for large-scale data storage cases, I think it will also cover the case that sharing data across multiple service nodes. Like log file is generated by jobservice node 1 but read and return by job service node2.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+
+Due to the  shorcomings of the above approaches, we explore the approach of using the registry as an intermediate store for system-generated data that needs to be shared across Harbor services. The next sections provide further details on this approach.
+
+### System Data Artifacts
+
+The proposal introduces the concept of a system data artifact. A system data artifact:
+* Is an arbitrary blob of data in text or binary.
+* Is created by a Harbor service to be consumed by other Harbor services.
+* Has a short lifespan - typically in minutes/hours but not in days.
+* Does not contain an OCI compatible manifest for describing the artifact data layers.
```
wy65701436, line 28: does the system artifact support replication? If so, how? 

Since the current replication framework only supports standard OCI artifact, can you add more details on that?

proposals/new/system-artifact-manager.md
steven-zou: I don't think short is a `MUST` criteria, it should be `data with living time`. May be several hours or several months. Those are all reasonable use cases.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+// logic involved in managing and tracking system artifacts.
+type SystemArtifactManager interface {
+
+	// Create a system artifact described by artifact record.
+	// The reader would be used to read from the underlying data artifact.
+	// Returns a system artifact tracking record id or any errors encountered in the data artifact upload process.
+	// Invoking this API would result in a repository being created with the specified name and digest within the registry.
+	Create(ctx context.Context, artifactRecord *SystemArtifact, reader io.Reader) (int64, error)
+
+	// Deletes a system artifact identified by a repository name and digest.
```
wy65701436, line 87: no data read() in the interface? and for the performance perspective, will it be supported cache?

proposals/new/system-artifact-manager.md
steven-zou: This is just a DAO model which is designed for DAO layer functions. So you need also to define a DAO interface to manage those table data (models).

proposals/new/system-artifact-manager.md
steven-zou: For the manager interface, we need another model to serve, different from the DAO model defined above.

```go
type SystemArtifact struct {
  ID int64 `json:"id"`
	// the name of repository associated with the artifact
	Repository string `json:"repository""`
	// the SHA-256 digest of the artifact data.
	Digest string `json:"digest""`
	// the size of the artifact data in bytes
	Size int64 `json:"size"`
	// the harbor subsystem that created the artifact
	Vendor string `json:"vendor"`
	// the type of the system artifact.
	// the type field specifies the type of artifact data and is useful when a harbor component generates more than one
	// kind of artifact. for e.g. a scan data export job could create a detailed CSV export data file as well
	// as an summary export file. here type could be set to "CSVDetail" and "ScanSummary"
	Type string `json:"type"`
	// the time of creation of the system artifact
	CreateTime time.Time `json:"create_time"`
	// optional extra attributes for the system artifact
	ExtraAttrs map[string]string `json:"extra_attrs"`
}
```

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+3. Store generated data artifact on a data volume mounted within the service container - This approach exposes the filesystem to the service. This will allow consumers to store data files using the standard filesystem APIs. However, within Kubernetes environments, Harbor mounts only Persistent volumes and these persistent volumes cannot be re-mounted in different services. Hence this option cannot be adopted.
+
+Due to the  shorcomings of the above approaches, we explore the approach of using the registry as an intermediate store for system-generated data that needs to be shared across Harbor services. The next sections provide further details on this approach.
+
+### System Data Artifacts
```
wy65701436, line 22: IMO, the defination of `System Artifacts` is a little confusion, it's called aritfact but stored as a blob. 

Artifact in Harbor corresponds to OCI artifact.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,112 @@
...
+	// Invoking this API would result in a repository being created with the specified name and digest within the registry.
+	Create(ctx context.Context, artifactRecord *SystemArtifact, reader io.Reader) (int64, error)
+
+	// Deletes a system artifact identified by a repository name and digest.
+	// Also deletes the tracking record from the underlying table.
+	Delete(ctx context.Context, string, digest string) error
+
+	// List all the system artifact records that match the criteria specified
+	// within the query.
+	List(ctx context.Context, query *q.Query) ([]*SystemArtifact, error)
```
steven-zou, line 93: How to read the data of the artifact? It seems we need to the method in the interface to return a data reader for the specified data.

steven-zou (commented):
Comments left

proposals/new/system-artifact-manager.md
prahaladdarkin: Updated the proposal with the details around why PVs cannot be used.

proposals/new/system-artifact-manager.md
prahaladdarkin: @wy65701436  Like @steven-zou   mentioned, the `SystemArtifact` is just a DAO record to track the created blobs. The publishing of system data storage to harbor admin would need to be enhanced to read the aggregate storage used by system artifacts. I will update the proposal to reflect the same.

proposals/new/system-artifact-manager.md
prahaladdarkin: Agreed @steven-zou .  This also implies that the creators of the system artifacts have a mechanism to control which system artifacts are deleted/cleaned-up periodically. I have updated the proposal to include a `CleanupCriteria` interface that would allow the consumers to determine  the logic for specifying the periodic cleanup Ids.

proposals/new/system-artifact-manager.md
prahaladdarkin: It will call the ``PushBlob` method from the `registry.Client` interface.

prahaladdarkin (commented):
Incorporated or responded to review comments

proposals/new/system-artifact-manager.md
prahaladdarkin: @wy65701436  I have added a section in the proposal describing how to to keep the generated artifacts safe from being cleaned up by Garbage Collection

tianon (commented):
Since these are intended to be pushed into the registry itself, don't they have to be part of a specific repository?  What's the plan for namespacing those such that they don't collide with anything users might already have created?

I'm guessing the implementation is just going to be "blobs" inside a special repository?

(Seems like these are probably details it might be good to cover in the proposal?)

community.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
+	"folders": [
+		{
+			"path": "."
+		}
+	],
+	"settings": {}
+}
```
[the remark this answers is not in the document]
prahaladdarkin: @tianon - that is right - these are just going to be **blobs** within a specific repository. The proposal does have a section that explicitly states that these artifacts are not OCI artifacts. However I will update the proposal to mention this fact

I will remove the `community.code-workspace` file.

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,234 @@
...
+
+	// Deletes a system artifact identified by a repository name and digest.
+	// Also deletes the tracking record from the underlying table.
+	Delete(ctx context.Context, string, digest string) error
+
+	// Exists checks for the existence of a system artifact identified by repository and digest.
+	// A system artifact is considered as in existence if both the following conditions are true:
+	// 1. There is a system artifact tracking record within the Harbor DB
+	// 2. There is a BLOB corresponding to the repository name and digest obtained from system artifact record.
+	Exists(ctx context.Context, repository string, digest string) (bool, error)
```
steven-zou, line 138: I think the function of this method equals to Get() returned an artifact or a NOT_FOUND error. @prahaladdarkin

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,234 @@
...
+	GetUntrackedArtifacts(ctx context.Context) ([]string, error)
+
+	// GetStaleArtifacts gets a list of system artifacts that have a tracking record within the
+	// database, but the artifact actually does not exist on the registry storage.
+	GetStaleArtifacts(ctx context.Context) ([]*SystemArtifact, error)
+
+	// GetStorageSize returns the total disk space used by the system artifacts stored in the registry.
+	GetStorageSize(ctx context.Context) (int64, error)
+
+	// Register a clean-up criteria for a specific vendor and artifact type combination.
```
steven-zou, line 155: All the registered criteria will be called in a clean-up cycle?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,234 @@
...
+
+1. Delete from registry store all system artifacts that are found in the registry but do not have a corresponding tracking record in the database 
+2. Delete all tracking records that are present in the database, but have no corresponding artifact within the registry store.
+
+#### Isolation from existing Harbor Garbage Collection
+
+The current Harbor Garbage collection logic uses the `Artifacts` and the `ArtifactThrash` tables to identify the registry artifacts to be garbage collected. The standard artifact upload mechanism ensures that the `Artifact` tables are populated when pushing OCI artifacts to Harbor. 
+The `SystemArtifactManager` on the other hand **does not** insert any records into the `Artfact` tables. It manages it's own table to track system artifacts. Hence the existing Harbor garbage collection job does not know about the created system artifacts and hence would not clean them up.
+
+### Storage Quota Management
```
steven-zou, line 215: If clean-up works well, it seems storage quota is not a mandatory requirement as this function is always used by system service not by users directly.

steven-zou (commented):
@prahaladdarkin add further comments

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,255 @@
...
+* Digest - the digest of the data blob present within the repository.
+
+The following rules apply:
+
+* A `Vendor` can create **multiple repositories**
+* A `Repository` is associated with a single `Vendor`
+* A `Repository` can have multiple `Type`s of data blobs
+* A `Digest` is associated with a single data blob.
+
+To prevent namespace clashes between user-created artifacts and system artifacts, all system artifacts would be created under a top level system namesapce called `HarborSystem`
```
wy65701436, line 60: If we choose HarborSystem as the namesapce for system artifact, do we need to modife the project API to avoid user using it? And what about if the `HarborSystem` is existing in user env?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,255 @@
...
+
+The below diagram depcits the storage hierarchy associated with system aritfacts ![System Artifacts Storage Hierarchy](../images/system-artifact/csv-export-system-artifact-manager.jpg)
+
+The complete system artifact repository name would hence be : `harborsystem/<vendor>/<repository-name>/<type>`
```
wy65701436, line 64: Can you give an example for the scan report of hello-world?

And for the storage hierarchy, can you add the hierarchy of the backend storage?
Like, if you push an hello-world into Harbor, the hierarchy should be:

```
/data/registry/
└── docker
    └── registry
        └── v2
            ├── blobs
            │   └── sha256
            │       ├── 1b
            │       │   └── 1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792
            │       │       └── data
            │       ├── b8
            │       │   └── b8dfde127a2919ff59ad3fd4a0776de178a555a76fff77a506e128aea3ed41e3
            │       │       └── data
            │       └── d1
            │           └── d1165f2212346b2bab48cb01c1e39ee8ad1be46b87873d9ca7a4e434980a7726
            │               └── data
            └── repositories
                └── library
                    └── hello-world
                        ├── _layers
                        │   └── sha256
                        │       ├── b8dfde127a2919ff59ad3fd4a0776de178a555a76fff77a506e128aea3ed41e3
                        │       │   └── link
                        │       └── d1165f2212346b2bab48cb01c1e39ee8ad1be46b87873d9ca7a4e434980a7726
                        │           └── link
                        ├── _manifests
                        │   ├── revisions
                        │   │   └── sha256
                        │   │       └── 1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792
                        │   │           └── link
                        │   └── tags
                        │       └── latest
                        │           ├── current
                        │           │   └── link
                        │           └── index
                        │               └── sha256
                        │                   └── 1b26826f602946860c279fce658f31050cff2c596583af237d971f4629b57792
                        │                       └── link
                        └── _uploads

29 directories, 8 files
```

What would like if I scaned the hello-world and generated the report?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,255 @@
...
+#### Isolation from existing Harbor Garbage Collection
+
+The current Harbor Garbage collection logic uses the `Artifacts` and the `ArtifactThrash` tables to identify the registry artifacts to be garbage collected. The standard artifact upload mechanism ensures that the `Artifact` tables are populated when pushing OCI artifacts to Harbor. 
+The `SystemArtifactManager` on the other hand **does not** insert any records into the `Artfact` tables. It manages it's own table to track system artifacts. Hence the existing Harbor garbage collection job does not know about the created system artifacts and hence would not clean them up.
+
+### Storage Quota Management
+
+In the first iteration of the framework, the quota for system artifact data is unlimited. There is no mechanism to specify a storage quota for different kinds of system artifacts. System artifact data quota can be specified at one of the following levels
+
+1. A quota for each vendor who generates a system artifact.
```
wy65701436, line 240: For the current design, the quota is based on project, I think we cannot support it. cc @heww

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,255 @@
...
+2. Delete all tracking records that are present in the database, but have no corresponding artifact within the registry store.
+
+#### Isolation from existing Harbor Garbage Collection
+
+The current Harbor Garbage collection logic uses the `Artifacts` and the `ArtifactThrash` tables to identify the registry artifacts to be garbage collected. The standard artifact upload mechanism ensures that the `Artifact` tables are populated when pushing OCI artifacts to Harbor. 
+The `SystemArtifactManager` on the other hand **does not** insert any records into the `Artfact` tables. It manages it's own table to track system artifacts. Hence the existing Harbor garbage collection job does not know about the created system artifacts and hence would not clean them up.
+
+### Storage Quota Management
+
+In the first iteration of the framework, the quota for system artifact data is unlimited. There is no mechanism to specify a storage quota for different kinds of system artifacts. System artifact data quota can be specified at one of the following levels
```
wy65701436, line 238: So, the first iteration, we do nothing on quota? And, the size of system artifacts will never reflact in the Harbor system storage usaged?

proposals/new/system-artifact-manager.md
prahaladdarkin: Ageed @wy65701436 . I have updated the proposal with the following:
* I have changed the name of the internal namespace to be something more uncommon for a user `sys_harbor_ns`. We would need to update the middleware for the `Project` API to validate that there are no projects being created with this name. changing the middleware would also ensure that such a namespace is not created using the UI too.

proposals/new/system-artifact-manager.md
prahaladdarkin: Done. I have added the proposed hierarchy tree structure images.

proposals/new/system-artifact-manager.md
prahaladdarkin: While in the first iteration, the quota is unlimited, we can still calculate using the system artifact tables the exact disk space being used by the system artifact blobs. I have already mentioned this detail within the section `UI/API visibility of system artifacts`

community.code-workspace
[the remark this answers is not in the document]
prahaladdarkin: Done. removed

proposals/new/system-artifact-manager.md
prahaladdarkin: Yes. The details are present within the section `Artifact Cleanup`

proposals/new/system-artifact-manager.md
prahaladdarkin: @steven-zou  - `Exists()` is more stringent than get in it's semantics and implementation. As has been detailed within the Go-Doc, a BLOB is considered as existing if **both** of the following conditions are true:
* There is a system artifact tracking record within the Harbor DB
* There is an actual BLOB in existence within the storage hierarchy.
The reason for providing such a method is to have the clients clear differentiate between situations where-in a BLOB that was found was actually a dangling BLOB (and hence non-existent) that should ideally have been cleaned up but was not (due to some error) versus a valid blob being found.

wy65701436 (approved):
generally, lgtm. Let's move on, thanks.

heww (approved):

prahaladdarkin:
@goharbor/all-maintainers can you please take a look and approve/ review vote for the proposal?

a-mccarthy:
@goharbor/all-maintainers bumping @prahaladdarkin's  message

> can you please take a look and approve/ review vote for the proposal?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,259 @@
...
+### Tracking System Data Artifact Creation
+
+The following structure would be used to track the creation of system data artifacts within Harbor
+```go
+// SystemArtifact represents a tracking record for each system artifact that has been
+// created within registry using the system artifact manager
+type SystemArtifact struct {
+	ID int64 `orm:"pk;auto;column(id)"`
+	// the name of repository associated with the artifact
+	Repository string `orm:"unique;column(repository)"`
```
tianon, line 79: I might be misunderstanding the ORM (apologies if so), but `repository` on an individual artifact can't really be "unique", can it?  This is going to be the full `sys_harbor_ns/<vendor>/<repository-name>/<type>` string, right?

Rather, the combination of `repository`+`digest` would be unique, right? (as noted in `TableUnique` below)

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,259 @@
...
+	Vendor string `orm:column(vendor)`
+	// the type of the system artifact.
+	// the type field specifies the type of artifact data and is useful when a harbor component generates more than one
+	// kind of artifact. for e.g. a scan data export job could create a detailed CSV export data file as well
+	// as an summary export file. here type could be set to "CSVDetail" and "ScanSummary"
+	Type string `orm:column(type)`
+	// the time of creation of the system artifact
+	CreateTime time.Time `orm:"column(create_time)"`
+	// optional extra attributes for the system artifact
+	ExtraAttrs string `orm:"column(extra_attrs)"`
```
tianon, line 94: Is this arbitrary JSON?

proposals/new/system-artifact-manager.md
```diff
@@ -0,0 +1,259 @@
...
+System data artifacts do not have corresponding Manifests and hence  *do not* conform to the OCI Artifact specification.
+
+### UI/API visibility of system artifacts
+
+System artrifacts are created by internal harbor components and it is not necessary that all generated artifact data be user facing. For e.g. a particular feature may run a job that generates raw data and a subsequent job that runs would aggregate the data generated by the first job. Hence 
+If an external user or an API client is allowed to interact with the artifact manager and perform CRUD on system artifacts it would tamper the integrity of the generated artifact and in some cases even interfere with functionality behavior. Additionally, exposing this data to the system admin in the UI or API would result in poor user experience and confusion, since internal details would be leaked to the end user who would have no idea on how these artifacts are generated, their lifetime and utilization. 
+
+Hence system artifacts would not be exposed in the UI or via an API.
+
+The storage occupied by the Harbor registry would also consider the total space occupied by all the system artifacts.
```
tianon, line 259: This seems kind of strange to me -- they won't be shown at all anywhere (and there won't even be an API for getting any kind of information about them), but they _will_ be reflected in the storage calculations.  If one of these happens to misbehave and keep artifacts longer than it's supposed to (or there becomes a use case for having long-lived system artifacts), won't that cause some amount of user confusion as to why the value is bigger for reasons they can't see?

To be clear, I think I'm arguing that maybe they _should_ show up _somewhere_ -- even if only just to say how much space they're taking up.

tianon (commented):
Overall, this seems good to me - just a few minor comments/questions. :+1:

proposals/new/system-artifact-manager.md
prahaladdarkin: yes this would be arbitrary JSON for now

proposals/new/system-artifact-manager.md
prahaladdarkin: @tianon  your concern is valid.  However, IMHO for now it can also be solved through some UX

For e.g. if we were to segregate on the UI by mentioning the following two values:
*  the space used by user created artifacts
* the space used by system artifacts (under an umbrella label say `Space used by system data`
then the user would have a mechanism of knowing what is consuming more space. The user will not know exactly which component/artifact but would know that it has something to do with harbor internals.

The main dilemma is on how to to surface these attributes to the end user when: 
* The user can do nothing about them. Like one natural direction for users to go is to ask if there can be some type of storage quota control on these artifacts.
* They do not really mean anything to the user - in fact the user won't even be aware that these actually get created.
Probably as mentioned in the proposal, when we were to implement storage quota management for the system artifacts we can provide mechanisms to perform an external clean-up or displaying additional details. 
Would like to know your views. on this strategy. @heww  @wy65701436  please provide your inputs too on this particular comment

heww: I agree with @prahaladdarkin that we can show the `Space used by system data` for the customers in the portal.

wy65701436: `Space used by system data` is fine for me, so we can have a way for UI to get it, right?

Vad1mo:
This might be important for this proposal.


From: https://github.com/opencontainers/tob/blob/main/proposals/wg-reference-types.md

> [](https://github.com/opencontainers/tob/blob/main/proposals/wg-reference-types.md#scope)As cloud native development continues to grow, there is increased interest in evolving registries to natively store, discover, and pull a graph of content associated with specific container images in a registry. Use cases for said associated artifacts include but are not limited to Software Bill of Materials (SBoM), security scan results, and signing. Having a native way to store and discover artifacts associated with other artifacts enables end-users to answer the question of: “What SBOMs or signatures are associated with this container image?”

wy65701436:
@prahaladdarkin this can be merged.

Vad1mo (approved):

PR #182, opened 2021-12-03
Add cleanup uploads proposal

stonezdj:
Provide a solution to cleanup the upload folders in registry storage.


proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,115 @@
...
+    uploadpurging:
+      enabled: true
+      age: 168h
+      interval: 24h
+      dryrun: false
+```
+
+But it is fixed schedule and could not be changed on demand, and could not run dry run to get the detail files that can be removed.
+
+## Proposal
```
wy65701436, line 27: so, we decide to clean the uploads in GC? Will you list another options?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,115 @@
...
+
+Enhance the _upload cleanup by adding the API(`/api/registry/purge`) to the registryctl container.
+
+Adding three configure items:
+
+   1. `purge_upload_on_gc` -- boolean to indicate if enable purge upload directory when GC is running.
+
+   1. `purge_upload_older_than`-- Remove files which are created before the specified hours.
```
wy65701436, line 35: if we want to expose this parameter, it should be have a minimum value to avoid conflict.

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,115 @@
...
+
+   1. `purge_upload_async` -- boolean to indicate if the purge process is running async, for dryrun, it always false, means it always run with sync mode. if true, then run the purge process in async mode.
```
wy65701436, line 37: when it set to true, it seems highly unlikely that the GC job will timeout. Can you add more details on timeout scenairo?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,115 @@
...
+
+If `purge_upload_on_gc` is true, the cleanup API is called by current garbage collection API. Files to be removed in _upload folders are listed in the GC dry-run report.
```
wy65701436, line 39: Given the purge API supports dry-run, if we leverage it in GC dry run, the GC will likely timeout on dry run. This is not a good experience for end user. I suggest that, we do not provide purge dry-run in GC.

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,115 @@
...
+	if gc.purgeUpload {
+		// TODO: run purge upload with specified purge parameter here
+	}
+  ...
+}  
+```
+
+## Open issues
+
+  1. Because the current dry run only returns the files need to removed, it doesn't return the total size of these files, it is impossible to display the total size of files can be removed.
```
wy65701436, line 113: if user triggers multiple GC with async purge, it will have serveral gorountines running to remove _uploads. We need more investigation on that, will they impact backend storage or  impact the usage(cpu/mem) of regsitryctl container?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+```
+But it need to do it manually, and it doesn't support s3, gcs object storage.
+
+## Proposal
+
+Enhance the _upload cleanup by adding the API(`/api/registry/purge`) to the registryctl container.
+
+Adding three configure items:
+
+   1. `purge_upload_on_gc` -- boolean to indicate if enable purge upload directory when GC is running.
```
reasonerjt, line 40: Is this going to be a setting on the system level or per GC job?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+   ```
+   The sample payload should be 
+   ```
+   {
+     older_than: 168,
+     async:     true,
+     log_out:    true
+   }
+   ```
+   The sample response should be a list of files need to removed when dry_run is true or aysnc is false.
```
reasonerjt, line 70: Is `dry_run` part of the request body?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Compatibility
+
+The current implementation should work on all supported backend image storages, includes local file system, NFS, s3 storage, azure storage and google storage.
+
+## Implementation
+
+1. Add API to registryctl to wrap the purge function in distribution.
+   ```
+    POST /api/registry/purge
```
reasonerjt, line 60: Does it really delete the blobs or it simply returns a list?  Please clarify.

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+   }
+   ```
+   The array size will be limited to 1000 rows, the rows exceed 1000 will be omitted, if no file need to remove, it is emtpy.
+
+2. Update the registryctl client to add function PurgeUpload.
+
+  ```
+	PurgeUpload(olderThan int64, dryRun, logOut, async bool) error
+  ```
+  If run the async is true, the PurgeUpload is running with async mode, to avoid run multiple purge operation in the same time, need to set a key purgeupload_running=true in redis, and set its expire time to 24h, and when the PurgeUpload complete or exit, it should delete the redis key purgeupload_running.
```
reasonerjt, line 87: When it's triggered via GC will async always be false?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+       "/data/registry/docker/registry/v2/repositories/library/lambda/_uploads/xxs33sxx"
+     ]
+   }
+   ```
+   The array size will be limited to 1000 rows, the rows exceed 1000 will be omitted, if no file need to remove, it is emtpy.
+
+2. Update the registryctl client to add function PurgeUpload.
+
+  ```
+	PurgeUpload(olderThan int64, dryRun, logOut, async bool) error
```
reasonerjt, line 85: How do you plan to implement `olderThan` to check the age of the blobs in `_upload` folder?

proposals/new/cleanup_upload_folder.md
```diff
@@ -0,0 +1,121 @@
...
+
+3. Add three options in the UI to the current Garbage Collection configure page.
+
+![gc_schedule](../images/cleanup-upload/gc_schedule.png)
+
+4. Purge upload folders with GC -- boolean option, when it is true, then the next two opitions are enabled.
+
+   1. Purge upload with age (hours) -- input with an integer value
+
+   1. Purge upload async -- boolean option
```
reasonerjt, line 97: when it's set to async, how does the result of purging impacts the status of GC job, how do we track the result of purging of `_upload` folder?

proposals/new/cleanup_upload_folder.md
stonezdj: per GC job

proposals/new/cleanup_upload_folder.md
stonezdj: As wangyan's suggested, the dry_run is not exposed to end user.

stonezdj:
After discussion, we decided to implement this feature in config.yml of the registry for simplicity. close this proposal.

PR #183, opened 2021-12-03
proposal: async update pull_time and pull_count

chlins:
Fix: https://github.com/goharbor/harbor/issues/15963

Signed-off-by: chlins <chenyuzh@vmware.com>


proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+
+* Reduce the number of database locks.
+
+* Fewer operations on the database update.
+
+* Improve the throughput of the harbor in the case of high concurrent pull.
+
+## Non-Goals
+
+* Fully accurate `pull_time` and `pull_count` of artifact.
```
wy65701436, line 30: We should doc somewhere that the pull_time & pull_count is not 100% accurate if we decide use this approch.

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+will raise two goroutines to update the artifact's `pull_time` and `pull_count`
+in the background for every `pull` request, and since it is a concurrent
+operation on a single record of the same table, problems such as goroutines
+piling up and database lock waiting can occur in the case of a high concurrent pull,
+this will also take up the number of connections to the database and thus reduce
+the performance of harbor.
+
+## Abstract
+
+Asynchronous update and merge the update operations of artifact's `pull_time` and `pull_count` to reduce
```
wy65701436, line 17: can user select the sync mode? sync or async?

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+Implement new functions for async update `pull_time` and `pull_count`
+periodically.
+
+```go
+
+var (
+    // maybe we can expose this variable to harbor configuration or environment variable.
+	asyncFlushDuration = 60 * time.Second
```
wy65701436, line 119: we should at least give a env to configurate it.

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+
+* Fewer operations on the database update.
+
+* Improve the throughput of the harbor in the case of high concurrent pull.
+
+## Non-Goals
+
+* Fully accurate `pull_time` and `pull_count` of artifact.
+
+* Immediate synchronized `pull_time` and `pull_count` of artifact after pulling.
```
wy65701436, line 32: Not Immediate sync pull time maybe impact the tag retation & GC. Since, tag retention &
GC leverages the pull time to judge candidates.

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Implementation
+
+Related code path: `src/controller/event/handler/internal/artifact.go`
+
+### Rationale
+
+* Cache the operation of pull into the memory.
```
wy65701436, line 40: do we also consider to use redis? And can we the comparison of advantages and disadvantages?
Mem & Redis.

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+
+* Merge the operations and only cache the final state.
```
wy65701436, line 42: How do we handle the multiple core update in HA scenario?

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: we can consider provide a flag to turn on sync/async mode, but 'sync' mode actually is only the logic sync if use goroutine to update it.

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: sure.

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: yes, i think we can do some comparison between them to decide which one to use.

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: every core instance keeps self data, independent of each other to do update operation to database.

proposals/async-update-artifact-pull_time-and-pull_count.md
heww: I think we don't need to support the sync mode. We can update these to volumes in DB immediately when the sync period time is zero.

proposals/async-update-artifact-pull_time-and-pull_count.md
```diff
@@ -0,0 +1,278 @@
...
+				if err != nil {
+					log.Infof("failed to list tags when to update pull time, %v", err)
+				} else {
+					if len(tags) != 0 {
+						tagID = tags[0].ID
+					}
+				}
+			}
+
+			if err := artifact.Ctl.UpdatePullTime(ctx, artifactID, tagID, time); err != nil {
```
heww, line 170: We may need to check that the UpdatePullTime uses the optimistic lock so that we only update the pull time when `time` is bigger than the pull time in db.

proposals/async-update-artifact-pull_time-and-pull_count.md
reasonerjt: storing the data in mem may cause issue when there are more than 1 instance of `core`?

reasonerjt: When the instance of core scaled in and out the data will be lost?

proposals/async-update-artifact-pull_time-and-pull_count.md
reasonerjt: We should make sure asyncFlushDuration is relatively small.

proposals/async-update-artifact-pull_time-and-pull_count.md
reasonerjt: Update in the manager or DAO level to make it default behavior?

wy65701436 (approved):
lgtm

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: set async flush time to zero to keep original logic.

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: Let's document this on the website.

proposals/async-update-artifact-pull_time-and-pull_count.md
chlins: We can put the behavior in the DAO level.

tianon:
This is slow specifically because it's updating an existing database record, right?

What if instead the code just did an append/insert to some "audit" type table for each event (which should be very fast, given the way the database journaling works) and then have a job that periodically (on a delayed trigger of some kind, maybe?) tallied those up for a combined update that does multiple pulls worth at once?

chlins:
> This is slow specifically because it's updating an existing database record, right?
> 
> What if instead the code just did an append/insert to some "audit" type table for each event (which should be very fast, given the way the database journaling works) and then have a job that periodically (on a delayed trigger of some kind, maybe?) tallied those up for a combined update that does multiple pulls worth at once?

@tianon Thanks your advice, but actually we already have the access logs in the `audit_log` table now which like you say,  but the table is too heavy, there are many issues about the audit log is too big or too slow, we also consider some ways to rotate this table, so from this table to retrieve access log and combine them maybe also bring up other performance issues.

wy65701436 (approved):
lgtm

heww (approved):

zyyw (approved):

PR #220, opened 2023-03-26
Add proposal for official CLI project for Harbor

akshatdalton:


proposals/new/cli.md
```diff
@@ -0,0 +1,130 @@
...
+
+        // projectApiClient is defined in the `project.go` file
+		response, err := projectApiClient.GetProject(cliContext, &params)
+		if err != nil {
+		    fmt.Println(err)
+            os.Exit(-1)
+        }
+
+		jsonStr := JsonMarshallWithMultipleTypes(response)
+		fmt.Println(jsonStr)
```
akshatdalton, line 117: Later we can improve the way we show the response, in this case, we can show it in the tabular form.

OrlinVasilev:
looks good so far :)

OrlinVasilev:
@Vad1mo @wy65701436 can you take a look please :)

Vad1mo (dismissed):
I think it would also make sense to point out that having a CLI that is capable of working with multiple registries. It should be possible to switch between different contexts like in docker or kubectl is.

scraly:
I'm adding my 2 centes about this proposal.

I think having an official CLI to be able to manipulate harbor is more than necessary at the moment :-).

A CLI that allows you to do the same thing as the API.
An up-to-date CLI.
A CLI with the same Developer Experience as kubectl and docker would be great.

I don't know much about the process to ensure that there is an official CLI for Harbor but I think that several people, a community, could work on the subject.

And as we know, the most difficult thing is not to create a CLI but to ensure that it remains up to date when the API is also updated and that it follows the same release cycle as the project core.

So my question is,

bupd:
closing this since we already have https://github.com/goharbor/harbor-cli as a subproject of harbor now.

PR #223, opened 2023-04-18
Added proposal for enabling full api access to robot accounts through ui

PaarthAgarwal:
Feature proposal for enabling full api access to robot accounts through ui.


sidewinder12s:
I do want to flag that multiple users have been reporting for more than 2 years that the actual RBAC evaluation logic between projects and system seem to be broken when users attempted to manually create robot accounts with both levels on the same Robot account. 

https://github.com/goharbor/harbor/issues/13653#issuecomment-1205438111

https://github.com/goharbor/harbor/issues/6784#issuecomment-1122631784

https://github.com/goharbor/harbor/issues/16159#issuecomment-1305214383

https://github.com/goharbor/harbor/issues/18730

Without resolving those issues exposing this in the GUI is likely to just appear broken to end users.

Vad1mo (approved):

PR #229, opened 2023-06-15
CDEvents proposal

afrittoli:
Proposal to implement CDEvents (cdevents.dev) notifications for Harbor.

This proposal is not 100% complete yet, the "Implementation Design" section is TBD, and there are a few open questions.


Vad1mo (dismissed):

Vad1mo (approved):

OrlinVasilev (approved):
looks great to me :)

OrlinVasilev:
@chlins can you take a look please

proposals/cdevents-support.md
```diff
@@ -0,0 +1,255 @@
...
+
+- Audit trail: Harbor is used to store artifacts as part of software development lifecycle (SDLC) workflows. The CDEvents produced by Harbor are stored along with those produced by the SCM tools used to manage the source code, the build and test tools used to build it and the deployment tools used to deploy the artifacts. All these events can be used to build an audit trail for artifacts stored in Harbor.
+
+- GitOps: tools like Argo CD and Flux can poll image registries and detect a new version or tag for an image associated with a deployment they manage. They react by pushing a PR to the SCM repository which in turn triggers a deployment of the new image version. Both tools could react to an event rather than polling. If such event was in CDEvents format they can implement support for one event type rather than integrating with every possible registry. Argo CD has an RFC open to implement support of CDEvents. The CDEvent community reached out to Flux as well to discuss CDEvents support. 
+
+## Proposal
+
+### Supported Events
+
+Harbor events support today are documented in the [webhook notification documentation][harbor-webhook].
```
chlins, line 45: Some events here are defined based on the functional scenarios of Harbor. CDEvents seem to have some differences compared to CloudEvents. CloudEvents only standardizes the structure of payload data without restricting event types, while CDEvents appear to have events associated with types. This means that some events in question may not be within the scope of CDEvents currently or in the future. Therefore, should we consider implementing restrictions on the frontend, allowing only specific event subscriptions for CDEvents? For example, currently, only PUSH events are supported.

afrittoli: Thanks @chlins - that's a great point. I think only allowing users to select available events via the UI sounds like a good option. I'm not very experienced with front-end development, so I could probably use some help there.
Another question I have related to this is whether Harbor provides users with an alternative way of configuring webhooks, i.e. not via the UI, some configuration file or API. If so, the restriction would have to be applied there too.

chlins: Yes, this is where it is not suitable at present. The original design of harbor only supports custom extensions to the payload data structure, but does not limit the event type. If only such restrictions are applied to CDEvents, it feels too specific.

afrittoli: Perhaps we could alter [`GetSupportedEventTypes`](https://github.com/goharbor/harbor/blob/adf80e921e17a329842214ffe18722452d516069/src/pkg/notification/notification.go#L152) to be [`PayloadFormat`](https://github.com/goharbor/harbor/blob/adf80e921e17a329842214ffe18722452d516069/src/pkg/notification/notification.go#L36) specific.

afrittoli: @chlins we plan to work to [expand the list of events in CDEvents](https://github.com/cdevents/spec/issues/143), but it may not match the list of events from Harbor.  To that end, I will propose to the CDEvents community to introduce support in the SDKs for custom types.

In the meanwhile, given that Habor supports configuring multiple webhooks, it seems acceptable that the CDEvents ones won't send all event types, as long as we make it clear to users that only a subset of events will be sent. Would this be enough for you to move forward? Please let me know if there is any more changes I shall make to let this proposal move forward.

Vad1mo (approved):

wy65701436 (changes requested):
It needs to discuss

afrittoli:
> It needs to discuss

Thanks @wy65701436 - I'm happy to discuss about this - what is your preferred forum for discussion? Would that be the Harbor working group or here on the proposal or else?

OrlinVasilev:
@wy65701436 @afrittoli did you manage to figure out what is needed

OrlinVasilev:
@afrittoli what's the progress of this one do you need anything from our side?

afrittoli:
> @afrittoli what's the progress of this one do you need anything from our side?

Hi @OrlinVasilev, thanks for asking. I've been working with the CDEvents community to add:

- more artifact events https://github.com/cdevents/spec/pull/172 - Done
- support for custom events https://hackmd.io/LftfRirGRbKuAcLg9pdOag - In progress

The two features will allow covering all existing Harbor events through CDEvents, as requested.

Feel free to review the custom events proposal https://hackmd.io/LftfRirGRbKuAcLg9pdOag - I used Harbor as a case study for that.

afrittoli:
> > @afrittoli what's the progress of this one do you need anything from our side?
> 
> Hi @OrlinVasilev, thanks for asking. I've been working with the CDEvents community to add:
> 
> * more artifact events [Add user field to artifact events and download and delete events cdevents/spec#172](https://github.com/cdevents/spec/pull/172) - Done
> * support for custom events https://hackmd.io/LftfRirGRbKuAcLg9pdOag - In progress
> 
> The two features will allow covering all existing Harbor events through CDEvents, as requested.
> 
> Feel free to review the custom events proposal https://hackmd.io/LftfRirGRbKuAcLg9pdOag - I used Harbor as a case study for that.

An update on this. Support for custom events was just merged in CDEvents https://github.com/cdevents/spec/tree/main/custom. We'll probably cut a release after KubeCon and then I will be able to update the proposal and implementation with support for all existing Harbor events as requested.

Vad1mo:
The [registry](https://github.com/cdevents/spec/blob/main/custom/registry.md): maintainers of the specifications of CDEvents custom events are encouraged to add their specification to the shared [registry](https://github.com/cdevents/spec/blob/main/custom/registry.md)

 @wy65701436 @chlins @OrlinVasilev can we agree upon supporting or not supporting this feature and let @afrittoli know before more time is wasted?

chlins:
Since Harbor already supports clouddevents, it may already meet the needs of most cloud-native users. For cdevents, it is more likely to be focused on the area of continuous deployment, and the Harbor community hasn't received too many requests for cdevents support yet, so I'm not sure if this is really valuable or or applicable scenarios to harbor users , so I think we can hold it for a while and wait for more user feedback, and at the same time we can monitor whether there are other artifact registries supports the cdevents.

PR #233, opened 2023-10-11
Add proposal for identity aware proxy authentication mode

shengjiang3:
Add a proposal for the Identity Aware proxy authentication mode. The identity aware proxy authentication mode is intended for Harbor to verify and use the identity token from an upstream identity proxy for the user identity.


Vad1mo (approved):
makes sense to me, IMO, there are many companies out there that are doing it.

I think what would help the proposal to get approval is:

- Highlight use cases where OIDC isn't possible
- Highlight use cases and integration options that would become possible with that. (the k8s one is good step, but I am sure there are more)

OrlinVasilev:
@shengjiang3 can you fix DCO please

stonezdj (changes requested):
Could you please add more detail for the implementation? such as the configuration item and give some brief introductions to the key change to implement it?

PR #234, opened 2023-11-22
Proposal for harbor cosign signature verification

karaguo:
To add a proposal for cosign signature verification in Harbor


Vad1mo (commented):
🥁
Thank you for your proposal, this looks indeed valuable. 

There are a few edge cases and considerations we need to consider.

Signatures and Image proxying and replication. The proposal should consider and support that use case.
Separation of Ownership and Operation should be reflected in the application configuration. In our user base, Harbor instance are set up by ops teams but used by devs or Platform teams. To take this into account, setting and configuration should be doable on the UI level, and not via env var as they are used for base application configuration.

Vad1mo:
@karaguo I would also like to invite you to present your proposal at the [next community meeting](https://github.com/goharbor/community/wiki/Harbor-Community-Meetings).

If that doesn't work, I'll suggest you can also do a 15 min recording, that we can watch during the community meeting

Vad1mo:
@karaguo, could you contribute to the development of this feature?

karaguo:
@Vad1mo Thanks for the comments! Sgtm! I can work on a brief demo recording and follow up with the review process to kick off

proposals/new/cosign-signature-verification.md
```diff
@@ -0,0 +1,177 @@
...
+  ]
+}
+```
+
+* Strictly verify signature with trusted CA list
+
+To iterate over all trust anchor entities, signatures are only verified upon the list. The artifact will pass verification if any trust anchor entity can verify the artifact
+
+## Non-Goals
+* Cosign strict signature verification status shown at UI
```
Vad1mo, line 177: Why not? We could have 3 types of Icons, Red ❌, Orange(Checked), Green (Checked)✅
The current green becomes the new orange checked, and the green checked is the new strict

proposals/new/cosign-signature-verification.md
```diff
@@ -0,0 +1,177 @@
...
+A mechanism to enable cosign strict verification is designed to support feature rollout. A global configuration can be set at harbor installation to determine whether the feature is visible to users. On the other hand, a harbor project level configuration allows users to enable or disable the feature at a harbor project level.  
+
+#### Option 1 
+The option is to use env variable in harbor core deployment. Harbor core environment variable can look as: 
+```go
+COSIGN_STRICT_VERIFICATION_ENABLED:true
+```
+
+#### Option 2  (preferred)
+Another option is to use project schema.
```
Vad1mo, line 44: Would it be possible to have a System wide CA setting that can be overwritten or extended with project-specific keys? Can you address that?

Vad1mo: Can you also explicitly mention this way can be used via the UI or API? It is not so clear.

proposals/new/cosign-signature-verification.md
```diff
@@ -0,0 +1,177 @@
...
+  * If verification fails, it returns 412 error code with message The image is not signed in Cosign or The image doesn't pass Cosign signature verification.
+
+## Design
+ 
+### Terminology 
+
+* Content trust cosign middleware: refers to the existing signature verification to check whether the corresponding signature exists or not.
+* Content trust strict cosign middleware: refers to the new proposed layer to strictly verify whether the manifest is signed, and signed by trusted entities. 
+
+### How to enable cosign strict verification
```
Vad1mo, line 34: For design, and flow verification, I would also suggest attaching some UI mockups of system-wide and project specific configuration options.

Vad1mo (changes requested):
Great progress! We are almost there. I am in favor of the proposal. Now it is about the outline and presentation of the UI/UX

wy65701436:
Generally, we don't plan to include the signature validation as Harbor's current goal is to focus on OCI artifact management, rather than becoming a comprehensive solution. And we can rely on other service or client to handle the validation.

PR #237, opened 2024-01-24
Use Harbor ourselves

Vad1mo:
Use of our Harbor registry to distribute our own container images. We should be the first to adopt our own technology. If we are not willing to utilize our own application for day-to-date use, don't expect others to do so.


slushysnowman:
I like the idea, but there probably need to be mitigating measures taken to ensure that if this Harbor instance goes down, that it can come back up right - if it's referencing itself as source of images, there are potential issues.

We're in the process of doing something similar at the moment, we're going to use our internal Harbor to store all our platform images, but we're going build in an automatic failover mechanism to pull from (in our case) ACR in the case where Harbor is down.

Maybe something to include here in some way/shape/form?

Vad1mo:
> I like the idea, but there probably need to be mitigating measures taken to ensure that if this Harbor instance goes down, that it can come back up right - if it's referencing itself as source of images, there are potential issues.
> 
> We're in the process of doing something similar at the moment, we're going to use our internal Harbor to store all our platform images, but we're going build in an automatic failover mechanism to pull from (in our case) ACR in the case where Harbor is down.
> 
> Maybe something to include here in some way/shape/form?

Yeah it makes sense to replicate to other registries eg. docker hub, or ghcr in case the outage takes to  longer to recover.

Vad1mo:
The current registry is available in https://registry.goharbor.io, you can already use it to pull Harbor images. We currently replicate images from Docker Hub.

proposals/use-our-own-registry.md
```diff
@@ -0,0 +1,67 @@
...
+-
+-
+
+# **Implementation**
+
+![Untitled](assets/aws-infa-harbor.png)
+
+Better understanding of the end users' experience
+
+## Phase I
```
wy65701436, line 56: Can we introduce Phase 0?

I believe we should aim for a graceful migration and use the test assets as a pilot to validate the stability and availability of the system for a period of time.

Vad1mo: done

proposals/use-our-own-registry.md
```diff
@@ -0,0 +1,75 @@
...
+Status: In Progress
+Created by: Vadim Bauer
+Created time: January 24, 2024 1:17 PM
+Last Edited: January 24, 2024 1:52 PM
+
+# **Proposal:Use Our Own Registry**
+
+# **Abstract**
+
+Use of our Harbor registry to distribute our own container images. We should be the first to adopt our own technology. If we are not willing to utilize our  own application for day-to-date use, don't expect others to do so.
```
thcdrt, line 12:
```suggestion
Use of our Harbor registry to distribute our own container images. We should be the first to adopt our own technology. If we are not willing to utilize our own application for day-to-date use, don't expect others to do so.
```

proposals/use-our-own-registry.md
```diff
@@ -0,0 +1,75 @@
...
+
+# **Background**
+
+Today, Harbor is the most popular and widely recognized full-fledged container registry on the market. Yet, we are not using our own product ourselfs. Hence the [dogfooding](https://www.projectmanagement.com/articles/217092/eat-your-own-dog-food#_=_) title
```
thcdrt, line 16:
```suggestion
Today, Harbor is the most popular and widely recognized full-fledged container registry on the market. Yet, we are not using our own product ourselfs. Hence the [dogfooding](https://www.projectmanagement.com/articles/217092/eat-your-own-dog-food#_=_) title.
```

thcdrt (approved):

thcdrt (approved):

OrlinVasilev (approved):

stonezdj:
Our nightly e2e UI and the github merge action based on registry.goharbor.io, in last month, we see lot of fail because of the registry.goharbor.io, it blocks our code merge sometimes. We have spent lots of effort on retry and diagnostics.

proposals/use-our-own-registry.md
```diff
@@ -0,0 +1,75 @@
...
+- Use registry to push test artifacts
+- Parallel operation next to Docker Hub.
+
+## Phase I
+
+- Dedicated IaC Repo containing the infrastructure
+- Since all images from Docker Hub are already in Harbor.
+  - Replace image references in Documentation
+  - Replace Image references in Compose
+  - Replace image references in Helm
```
wy65701436, line 69: We aim to use this instance as the single source of truth for serving our release images, meaning all network traffic will be redirected from Docker Hub to this instance. The main concern is whether we can provide and ensure the same level of stability as Docker Hub, as well as allocate dedicated resources to monitor and maintain this instance.

Could we set a criterion where, if we don't encounter continuous failures while pulling images in the Harbor CI for three months, we can then consider switching the online package traffic to this instance and eventually to Harbor Helm?

Vad1mo: We can replicate images from registry.goharbor.io to dockerhub as a fallback, for cases if users can't pull from registry.goharbor.io

image

bupd (approved):
lgtm

wy65701436 (approved):
lgtm

PR #242, opened 2024-06-12
RFP for Multiple project levels

Vad1mo:
This proposal extends Harbor with additional project visibility and permission levels and upgrades the current public/private project visibility and access, making Harbor Suitable for more use cases.


OrlinVasilev (approved):

OrlinVasilev:
+1

snowmanstark (approved):

peter-englmaier:
Perhaps you want to have a way to explicitly tag users as "internal". So internal projects are not accidentally exposed to customers, who have an account to download software delivered to them. Otherwise, I find the proposal quite useful for larger organisations.

Vad1mo:
> Perhaps you want to have a way to explicitly tag users as "internal". So internal projects are not accidentally exposed to customers, who have an account to download software delivered to them. Otherwise, I find the proposal quite useful for larger organisations.

This is not in scope for this proposal. In this context, internal does not refer to users but to networks, this would be a separate use case IMO.

proposals/multiple-project-visibility-levels.md
```diff
@@ -0,0 +1,88 @@
...
+Internal projects are a bit special in terms of internal visibility accessibility.
+On the one hand,
+we want the internal projects
+to be visible and accessible inside the organization, but private to the outside.
+The same would apply to pulling artifacts,
+internally without a pull secret and externally just like a private project with a pull secret.
+
+To handle this the internal use case, we would need
+to have a functionality
+that can distinguish between access from internal or external networks.
```
tillepille, line 79: I think adding the complexity of networking here makes the proposal bigger than it needs to be. 

For me it's about authorizing all authenticated identities to a project by default.

IMHO adding networking brings much complexity and risk with it.

proposals/multiple-project-visibility-levels.md
```diff
@@ -0,0 +1,88 @@
...
+
+Extend the current public and private project visibility levels with additional options.
+
+- **Private (Existing)**
+  Same behavior as the current private project visibility.
+  Only users and groups
+  that are explicitly added as members to the project to get access,
+  based on their access level.
+- **Internal View Only**
+  All authenticated (hence authenticated)
```
tillepille, line 43: "hence authenticated", you mean authorized?

tillepille (commented):
👍

proposals/multiple-project-visibility-levels.md
```diff
@@ -0,0 +1,89 @@
+# Proposal: Multiple Visibility Levels and Control
+
+Author: [Vadim Bauer](https://github.com/Vad1mo) - [8gears Container Registry](https://container-registry.com/)
+Created time: 11 June 2024
+Last updated:
+
+## Abstract
+
+This proposal extends Harbor with additional project visibility and permission levels, and upgrades the current public/private project visibility and access levels.
```
wy65701436, line 9: Generally, the document lacks of technical parts to let other maintainers know how to handle the newly introduced level? 

> The scheme changes? And the data migration?
> What the RBAC mode?
> API change?
> Any downstream impact?

And also, could you add persona? user story?

proposals/multiple-project-visibility-levels.md
strigazi: I would leave this out of scope of this proposal, maybe it can become an appendix in the documentation. For info, at CERN we have configured an allow-list with our internal subnets in the ingress.

Vad1mo:
It make sense that all robot accounts (project/system) should have automatically access to all internal projects.

PR #250, opened 2024-08-19
Add proposal CLI-secret-for-api

LGhoull:


LGhoull:
This is a kind/proposal, but I can't add any labels

proposals/new/CLI-secret-for-api
```diff
@@ -0,0 +1,46 @@
...
+
+## Non-Goals
+
+-
+
+## Rationale
+
+This change can be seen as a security "downgrade", but since this way of authenticating is already present when using the docker cli, the vulnerability already exists (if it can be considered one).
+
+This change disables the authentication via id token. Alternatively, one could allow both ways of authentication at the same time and only deprecate the id-token-way, but since the id token is such a "bumpy" solution (in my opinion) anyway, I don't think it is necessary.
```
reasonerjt, line 34: I suggest we deprecate it and remove it in future release, otherwise, this MAY break some users existing workflow.

proposals/new/CLI-secret-for-api
```diff
@@ -0,0 +1,46 @@
...
+- remove idtoken security context generator
+- rename and consolidate tests and names accordingly
+
+## Non-Goals
+
+-
+
+## Rationale
+
+This change can be seen as a security "downgrade", but since this way of authenticating is already present when using the docker cli, the vulnerability already exists (if it can be considered one).
```
reasonerjt, line 32: We should double check  if the token is invalid the secret becomes invalid at the same time.

proposals/new/CLI-secret-for-api
```diff
@@ -0,0 +1,46 @@
...
+The usual workaround for this is to use robot accounts, but that would mean that each user has to own a separate robot account just to access the API, which doesn't seem clean to me.
+Also, this means that robot accounts can not be created via the API by someone other than the local admin.
+
+The OIDC id token is accessible via either the harbor debug log, which the user can't see, or via direct communication with the identity provider, which requires credentials that the user isn't supposed to have.
+
+## Proposal
+
+- Extend oidc_cli security context generator to include calls to the v2 API
+- remove idtoken security context generator
+- rename and consolidate tests and names accordingly
```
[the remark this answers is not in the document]
LGhoull: You mean the id token? From my understanding, harbor attempts to renew the id token to keep the secret valid. If that's not possible, the Secret becomes invalid. All this logic already exists for the CLI secret

proposals/new/CLI-secret-for-api
LGhoull: What I meant with this being a possible "vulnerability" (depending on the way you look at it, personally I don't think it is a vulnerability) is the fact that the CLI secret only needs to be copied from the UI once and can be used repeatedly without signing in with OIDC again. But:

- This is already the case for the CLI secret
- The CLI secret is invalidated when the id token can not be renewed (for example when the user in the ID provider is deleted)
- This is basically how API keys work in numerous other applications

LGhoull:
@reasonerjt Could you respond to the discussion?

proposals/new/CLI-secret-for-api
```diff
@@ -0,0 +1,44 @@
+# Proposal: Use CLI secret for auth against the API (OIDC)
+
+Author: Luis Fittkau
+
+Discussion: -
+
+## Abstract
+
+Allow basic auth, consisting of username and CLI secret, in addition to the OIDC id token when authenticating against the API when OIDC is enabled. Deprecate auth via OIDC id token.
```
Vad1mo, line 9: Can you outline the backward compatibility or its lack of?

proposals/new/CLI-secret-for-api
```diff
@@ -0,0 +1,44 @@
...
+
+The OIDC id token is accessible via either the harbor debug log, which the user can't see, or via direct communication with the identity provider, which requires credentials that the user isn't supposed to have.
+
+## Proposal
+
+- Extend oidc_cli security context generator to include calls to the v2 API
+
+## Non-Goals
+
+-
```
Vad1mo, line 26: What about `docker login `

Vad1mo (changes requested):
It is very difficult for me to follow on the details of the PR, on high level I think  I understand what you are trying to achieve and why. 

There are lot of details missing the description. Especially, all the backward-compatible and changes neeed to UI, Docs, and workflows are missing.

proposals/new/CLI-secret-for-api
LGhoull: docker login already works with the cli secret and nothing changes about that

LGhoull:
@Vad1mo I just added some clarification in the proposal. Of course we should document the new behavior. The UI doesn't have to change since CLI secrets already exist (maybe we can rename them).

proposals/new/CLI-secret-for-api
LGhoull: just added some more clarification in the compatibility section. The id token can still be used, but is deprecated.

LGhoull:
Can we move forward here please?

VinuelMancent:
Can I assist on getting this PR merged? The change is IMHO very important, but since nothing happened here for months, i fear this will never see the light of day

LGhoull:
@Vad1mo @chlins @zyyw Can we move forward here please?

malmor:
Hey there, any update on this proposal?

In https://github.com/goharbor/harbor/issues/16647#issuecomment-2538227742 a different concept of `classic API keys` was mentioned. Should this be re-written (or replaced) by that idea?

malmor:
Hey @Vad1mo @reasonerjt,
just wanted to check in on this proposal. Is there anything we can do to support this in moving forward?

Vad1mo:
the proposal and PR introduces a breaking change, this needs to be addressed and resolved, otherwise it has no chance getting in.

PR #253, opened 2024-11-28
Enhance audit log

stonezdj:


proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+Harbor is an important component in the cloud-native software security supply chain, it is used to store and distribute artifact through the software develement lifecycle. many users concern about the audit log to log important security events in the Harbor, so that every security incident can be traceable. They want to enrich the event type to add the following:
+
+1. User login, include success and failed login
+1. User create/delete
+1. Project member add/remove, including user and group member
+1. Configuration change, including system-level config and project-level configure
+1. Project policy change: include the tag retention, and immutable policy change
+1. Audit log cleanup schedule or execution
+
+With the above event type, the Harbor administrator can trace the user's behavior in the system, and know who has done what in the system, and when it happened, and the source of the request. make the Harbor more secure and traceable.
```
reasonerjt, line 20: I suggest we first clarify "audit log".  Is it only for tracing "user triggered" event?
For example if a scan is triggered by schedule, should it be traced as "audit log"?

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+    ProjectID    int64     `orm:"column(project_id)" json:"project_id"`
+    Operation    string    `orm:"column(operation)" json:"operation"`
+    OperationDescripton string
+    OperationResult string
+    ResourceType string    `orm:"column(resource_type)"  json:"resource_type"`
+    Resource     string    `orm:"column(resource)" json:"resource"`
+    Username     string    `orm:"column(username)"  json:"username"`
+    OpTime       time.Time `orm:"column(op_time)" json:"op_time" sort:"default:desc"`
+    RequestPayload string
+    IPAddress    string
```
reasonerjt, line 61: As discussed in another PR.  The IP address can be misleading in many ways, are you sure we wanna add it?

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+## Event Format
+The audit log format should be changed as follow
+
+```go
+type AuditLog struct {
+    ID           int64     `orm:"pk;auto;column(id)" json:"id"`
+    ProjectID    int64     `orm:"column(project_id)" json:"project_id"`
+    Operation    string    `orm:"column(operation)" json:"operation"`
+    OperationDescripton string
+    OperationResult string
```
reasonerjt, line 55: Is it string or bool?

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+    ResourceType string    `orm:"column(resource_type)"  json:"resource_type"`
+    Resource     string    `orm:"column(resource)" json:"resource"`
+    Username     string    `orm:"column(username)"  json:"username"`
+    OpTime       time.Time `orm:"column(op_time)" json:"op_time" sort:"default:desc"`
+    RequestPayload string
```
reasonerjt, line 60: IMO we should add a translation layer to translate the request body to event-specific parameters.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
+Proposal: Enhance Audit Log
```
reasonerjt, line 1: Let's highilght in the title that this is WIP or draft, if it's not ready for review by the community.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Related UI changes
+
+### Disable Audit Log Event Type
+
+Add a new configuration item in the system configuration page to disable the unwanted audit log event type.
+Add the `audit_log_disable` configuration item in the Configuration -> System Settings page, the user can input the event type to disable the audit log event, the event type should be separated by comma. Because the event type is the combination of operation and resource type, the user can input the operation and resource type to disable the audit log event. it is a string type configuration item. 
+
+![Disable Audit Log Event Type](../images/enhance_auditlog/config_disable_event_type.png)
+
```
reasonerjt, line 257: 1. I think we also need to explain supporting different backends for storing audit log, and how to make it possible to send to other backends like open-telemetry?
2. Are there any consideration in Querying the audit-log?  Are there any changes needed in the API?   Will we only support querying audit-log in Harbor API when the backend of audit-log is DB?

Vad1mo:
This proposal not only introduces a new auditlogo but also add much more events.

- The volume of transactions that the new system can cause in a high pull/push environments. One crucial part of the enhanced auditlog should be the decoupling of event producer action and  event writing.  In other words, an artifact pull/push should not be affects by the eventlog.

- The Database is currently the bottelneck of Harbor, especially on large installations, with many artifacts, polcies, voulnerability data. Reducing the pressure on the Database should be priority in the new auditlogs. Adding Redis here will not improve the situation, unless we use/implement a Write-Back Caching strategy.

I suggest also making the auditlog plugable, meaning that in the future it would be possible to add other types of backends apart from postgres, like opentelemtry or loki..

proposals/new/enhance_audit_log.md
wy65701436, line 61: should we record the ip address? As I recall that we have concern the ip may not reflect the truth.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+		}
+        // use a wrapper to get the response code and response header
+		rw := &ResponseWriter{
+			ResponseWriter: w,
+			statusCode:     http.StatusOK,
+		}
+
+		next.ServeHTTP(rw, r)
+
+		if enableAudit {
```
wy65701436, line 151: do we record non-200 actions? like someone attempted to access a forbidden resource or login failed.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+
+Harbor is an important component in the cloud-native software security supply chain, it is used to store and distribute artifact through the software develement lifecycle. many users concern about the audit log to log important security events in the Harbor, so that every security incident can be traceable. They want to enrich the event type to add the following:
+
+1. User login, include success and failed login
+1. User create/delete
+1. Project member add/remove, including user and group member
+1. Configuration change, including system-level config and project-level configure
+1. Project policy change: include the tag retention, and immutable policy change
+1. Audit log cleanup schedule or execution
+
```
wy65701436, line 19: I think, we also need to consider that the Queryable. Like, an administrator want to filter out all logs where user A performed action B from source C between the dates D and E.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+1. As a Harbor administrator, I want to know who has changed the project configuration, so that I can trace the user's behavior in the system.
+1. As a Harbor administrator, I want to know who has changed the project policy, so that I can trace the user's behavior in the system.
+
+# Solution
+
+## Event Format
+The audit log format should be changed as follow
+
+```go
+type AuditLog struct {
```
wy65701436, line 50: I'm not sure how to ensure immutability from a database perspective.

stonezdj:
> This proposal not only introduces a new auditlogo but also add much more events.
> 
> * The volume of transactions that the new system can cause in a high pull/push environments. One crucial part of the enhanced auditlog should be the decoupling of event producer action and  event writing.  In other words, an artifact pull/push should not be affects by the eventlog.
> * The Database is currently the bottelneck of Harbor, especially on large installations, with many artifacts, polcies, voulnerability data. Reducing the pressure on the Database should be priority in the new auditlogs. Adding Redis here will not improve the situation, unless we use/implement a Write-Back Caching strategy.
> 
> I suggest also making the auditlog plugable, meaning that in the future it would be possible to add other types of backends apart from postgres, like opentelemtry or loki..
> 
> I am looking into few options and [Unlogged tables ](https://pganalyze.com/blog/5mins-postgres-unlogged-tables) directly or in combination with stored procedures can provide a huge improvement in write performance.

Previous audit log has the log forward option, it can avoid log in the database
Screenshot 2024-12-11 at 16 39 55

proposals/new/enhance_audit_log.md
stonezdj: correct

proposals/new/enhance_audit_log.md
stonezdj: We can use the q.Query parameter to extend the queryable on audit log.

Vad1mo:
> Previous audit log has the log forward option, it can avoid log in the database

The problem with the existing approach is that when this is enabled. Audit logs aren't displayed in Harbor anymore. All events are forwarded to syslog. 

What I had in mind was had the option to implement a different storage backend instead of the postgres.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,440 @@
...
+
+### Cleanup Audit Log
+
+In the previous implementation, only image related event types could be selected to purging, such as create/delete/pull. we need to add more options to select new event types. Because there are too many event types to display in the UI, just provide the resource type to clean up the audit log. The cleanup operation will also apply to the legacy audit log table.
+
+
+![Cleanup Audit Log Event Type](../images/enhance_auditlog/cleanup_audit_log.png)
+
+
+## Schema Change
```
wy65701436, line 376: please also include the audit v2 apis in this proposal.

stonezdj:
> > Previous audit log has the log forward option, it can avoid log in the database
> 
> The problem with the existing approach is that when this is enabled. Audit logs aren't displayed in Harbor anymore. All events are forwarded to syslog.
> 
> What I had in mind was had the option to implement a different storage backend instead of the postgres.

When you configured forward log to LogInsight or ELK, you should search the audit log in the ELK's query interface.

reasonerjt:
> The problem with the existing approach is that when this is enabled. Audit logs aren't displayed in Harbor anymore. All events are forwarded to syslog.
> 
> What I had in mind was had the option to implement a different storage backend instead of the postgres.

@Vad1mo This is essentially a query module in Harbor that works with different backends.
The requirement to provide a pluggable query module in Harbor for audit logs should be tracked separately. 
And as Stone has suggested, when the user chooses to store Harbor's audit logs to a 3rd party system, normally he will query the information from the 3rd party system instead from Harbor.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+## Non-Goals
+
+1. The current audit log is based on the http middleware, it means it can only capture the event has http request and http response, and it is initiated by the user, usually it is a user action. for system level background job, such as the job service, the event is not captured by the audit log. it is out of the scope of this proposal.
+
+1. User login event only includes logins through browser, it does not include basic authentication by Docker CLI such as `docker login`, nor does it include basic authentication when calling  RESTful APIs. 
+
+## Terms
+
+
+
```
reasonerjt, line 553: Some overall suggestions:
1.  Clearly mention what will be implemented in v2.13
2. Add a few paragraphs to address common issues, like "source IP", support sending audit logs to 3rd party service other than Harbor's DB.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+```go
+type AuditLogV2 struct {
+    ID           int64     `orm:"pk;auto;column(id)" json:"id"`
+    ProjectID    int64     `orm:"column(project_id)" json:"project_id"`
+    Operation    string    `orm:"column(operation)" json:"operation"`
+    OperationDescripton string
+    OperationResult string
+    ResourceType string    `orm:"column(resource_type)"  json:"resource_type"`
+    Resource     string    `orm:"column(resource)" json:"resource"`
```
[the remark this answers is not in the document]
stonezdj: Prefer resource name, some resource has no name, it will be ID.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+1. Add new event type in the controller/event/topic.go file, and implement the ResolveToAuditLog method to resolve the current event to the audit log.
+1. Add topic for each event type in the controller/event/topic.go file, and register the event handler in the controller/event/handler.go file.
+1. Update the controller/event/handler/init.go file to add the auditlogHandler to subscribe to the new event type.
+1. Update the controller/event/handler/auditlog/auditlog.go file to handle the new event type.
+
+The above steps are cumbersome and error-prone when there are too many event types. To simplify the process, we can create a common event metadata and a common event that can handle all events related to the v2.0 API. When an event occurs, create a metadata and call notification.AddEvent(ctx, event, true) to send the event queue. The middleware will process the event in the queue when the request is complete. These events will be handled asynchronously. These events will be sent to same topic, the audit log handler will fetch the event from the queue, call ResolveToAuditLog, and create an audit log item.
```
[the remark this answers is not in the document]
stonezdj: current is used for audit log handler, but it can be subscribed by other handler

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+
+```go
+
+func Middleware() func(http.Handler) http.Handler {
+    ...
+    // Add audit log middleware
+		enableAudit := false
+		urlStr := r.URL.String()
+		username := "unknown"
+		re := regexp.MustCompile("^/c/log_out$")
```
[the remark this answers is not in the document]
stonezdj: `GET /c/log_out` is used to check if it is a logout event.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+
+AuditLog --|> Database : default
+AuditLog --|> LogStash : forward
+AuditLog --|> LogInsight : forward
+```
+
+## Disable Specific Audit Log Event 
+Because there are lots of event types add to the audit log, some user might need to skip some unwanted event type. to disable it, the user need to configure audit_log_disable like that
+```
+create_user, delete_user, update_user
```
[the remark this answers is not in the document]
stonezdj: just keep the configuration item as simple as possible

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+		p := regexp.MustCompile(url)
+		if p.MatchString(c.RequestURL) {
+			return r.Resolve(c, event)
+		}
+	}
+	return nil
+}
+
+```
+For each basic event type, we can create a resolver to resolve the common event to the specific event type, such as userResolver, projectResolver, robotResolver, etc.
```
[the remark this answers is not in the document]
stonezdj: OK, update it

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+	}
+	return false, ""
+}
+```
+
+In the handler of audit log, add condition to check if the current <operation>_<resourcetype> is enabled, default value is enabled for all event type.
+```go
+func (h *Handler) Handle(ctx context.Context, value interface{}) error {
+	...
+		if auditLog != nil && config.AuditLogEnabled(ctx, fmt.Sprintf("%v_%v", auditLog.Operation, auditLog.ResourceType)) {
```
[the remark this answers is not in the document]
stonezdj: Because the previous audit_log_disable should apply to the existing audit log implementation, such as image create image delete, it should be check here.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+		rw := &ResponseWriter{
+			ResponseWriter: w,
+			statusCode:     http.StatusOK,
+		}
+
+		next.ServeHTTP(rw, r)
+
+		if enableAudit {
+			ctx := r.Context()
+			event := &commonevent.Metadata{
```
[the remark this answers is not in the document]
stonezdj: the function PreCheck() can filter the event before propagating.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,555 @@
...
+			// If this event is not captured, we don't need to add it to the notification
+			if !event.PreCheck(ctx) {
+				return
+			}
+```
+
+The Metadata's PreCheck method will match the url pattern and event resolver to check if current event is disabled. if it is disabled, it will skip calling SendEvent() method to save the subsequent overhead. the following is the code snippet of the PreCheck method.
+
+```go
+func (c *Metadata) PreCheck(ctx context.Context) (bool, string) {
```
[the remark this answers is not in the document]
stonezdj: filter event before propogating

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,577 @@
...
+- https://github.com/goharbor/harbor/issues/20292
+- https://github.com/goharbor/harbor/issues/18351
+- https://github.com/goharbor/harbor/issues/15134
+- https://github.com/goharbor/harbor/issues/14277
+- https://github.com/goharbor/harbor/issues/4426
+- https://github.com/goharbor/harbor/issues/11996
+
+# Personas and User Stories
+
+1. As a Harbor administrator, I want to know who has logged in/logout to the system, so that I can trace the user's behavior in the system.
```
wy65701436, line 36: As discussed, we prefer to do the 1,3 and 4. Please update it accordingly.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,577 @@
...
+        '500':
+          $ref: '#/responses/500'
+```
+and it returns the `AuditLogEventType`
+
+```
+  AuditLogEventType:
+    type: object
+    properties: 
+      event_id:
```
reasonerjt, line 474: Isn't the event_id and name a dup?  We'll need to transform the response to the label for display either way.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,577 @@
...
+```
+With this feature, the previous configure option `pull_audit_log_disable` can be deprecated. Just use `audit_log_disable` to disable the unwanted event type.
+
+## Cleanup Audit Log
+
+Because purge audit log will delete the audit log periodically,  and it allow user to select the event to purge, the new type of event should be added to the selection, such as user login/logout, user create/delete, project member add/remove, configuration change, project policy change. it involves too many event type, so we can categorize all these events as `other events` type. 
+
+## API change
+
+Add a new API to query the audit log ext event. the API is `/api/v2.0/auditlog-exts`, the user can filter the audit log event by the operation, resource_type, resource, the operation description and operation result are visible to users.
```
reasonerjt, line 329: This is trivial but `auditlogs-ext` is better b/c the ext is just a suffix to distinguish it from the old resource.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,574 @@
...
+        type: string
+        format: date-time
+        example: '2006-01-02T15:04:05Z'
+        description: The time when this operation is triggered.
+```
+
+In order to provide a list of audit log event types to disable in configuration pannel, the API `/api/v2.0/auditlog-exts/events` provide the event types ID and names.
+
+```
+  /auditlog-exts/events:
```
Vad1mo, line 444: what does the **exts** mean in the api? 

Why not make it clear to the user eg as in **auditlog2**

Vad1mo (changes requested):

proposals/new/enhance_audit_log.md
stonezdj: It is just a extension of the audit log, add ext suffix.

proposals/new/enhance_audit_log.md
```diff
@@ -0,0 +1,299 @@
...
+
+# Solution
+
+## Event Format
+The audit log format should be changed as follow
+
+```go
+type AuditLog struct {
+    ID           int64     `orm:"pk;auto;column(id)" json:"id"`
+    ProjectID    int64     `orm:"column(project_id)" json:"project_id"`
```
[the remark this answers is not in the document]
stonezdj: The project id would be empty, the project id is required when querying audit log with project, because audit log can query under project tab.

proposals/new/enhance_audit_log.md
stonezdj: Updated

proposals/new/enhance_audit_log.md
stonezdj: Keep the IPAddress as a reserved column

proposals/new/enhance_audit_log.md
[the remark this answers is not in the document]
stonezdj: payload is a reserved column for future usage.

proposals/new/enhance_audit_log.md
stonezdj: Already updated in the proposal

stonezdj:
> > Previous audit log has the log forward option, it can avoid log in the database
> 
> The problem with the existing approach is that when this is enabled. Audit logs aren't displayed in Harbor anymore. All events are forwarded to syslog.
> 
> What I had in mind was had the option to implement a different storage backend instead of the postgres.

Replace postgres is a big task, we have no bandwidth to handle so far.

chlins (approved):
lgtm

reasonerjt (approved):

PR #256, opened 2025-01-06
Add Single Active Replication Proposal

bupd:

### DIFF
--- a/proposals/new/single-active-replication.md
+++ b/proposals/new/single-active-replication.md
@@ -1,0 +1,118 @@
+# Proposal: Single Active Replication
+
+Author: Prasanth Baskar/[bupd](https://github.com/bupd)
+
+Discussion & PR: [https://github.com/goharbor/harbor/pull/21347](https://github.com/goharbor/harbor/pull/21347)
+
+## Abstract
+
+This proposal introduces a new feature that adds and option to prevent the parallel execution of replications in a given replication policy. By adding a "Single Active Replication" checkbox in the replication policy, users can ensure that only one replication task for the same replication policy is executed at a time, preventing unnecessary resource consumption, reducing bandwidth throttling, and improving overall replication performance.
+
+## Problem
+
+Currently, Harbor allows multiple replication executions of the same policy to run in parallel. When policies are scheduled frequently and involve large artifacts (e.g., 1 GB+), overlapping executions may:
+- Copy the same big artifact layers
+- Consume unnecessary bandwidth and IOPS
+- Cause throttling or failures in low-bandwidth environments
+- Result in exponential performance degradation
+This issue is exacerbated when artifact sizes increase and scheduling intervals are frequent.
+
+## User Stories
+### Story 1
+As a user with Harbor deployed across multiple regions, I rely on scheduled replication policies to keep them synchronized. The constant problem I face is that these scheduled replications often overlap, leading to the same policy being run simultaneously. I desperately need replications for a given policy to run sequentially, not concurrently.
+
+### Story 2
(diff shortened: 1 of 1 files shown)


Vad1mo (approved):

thcdrt (approved):

OrlinVasilev (approved):

reasonerjt:
@bupd Thanks for the proposal, I think this is a valid scenario and the design in UX makes sense to me.

However, in this design, it lacks the details in terms of how to implement it in the backend.  

I think more details are needed like:
-  Will we need a global lock for harbor-core, or jobservice?
-  How to ensure the lock is functional when there are multiple instance of harbor-core?
-  What is the data structure of this lock?
-  The code in which file or package needs to be updated to trigger the "lock" "unlock" action.
- How does it fail-over, for example if a lock is added and failed to unlock, what's the impact and how to mitigate it.

reasonerjt (changes requested):
This is a valid feature enhancement but this design needs more details.

bupd:
> I think more details are needed like:
Will we need a global lock for harbor-core, or jobservice?

No, we are checking the currently running replications from DB and that does not require locks.

> How to ensure the lock is functional when there are multiple instance of harbor-core?

Invalid since there are no locks present.

> What is the data structure of this lock?
> The code in which file or package needs to be updated to trigger the "lock" "unlock" action.
> How does it fail-over, for example if a lock is added and failed to unlock, what's the impact and how to mitigate it.


The above are **Invalid**, Since there is no lock happening on DB, core or jobservice. we are just checking the current replication executions and check if the previous replication executions are completed before starting a new one. 

Thanks @reasonerjt  I will update proposal. clearly stating that no locks/unlocks are involved.

reasonerjt:
> No, we are checking the currently running replications from DB and that does not require locks.
I don't quite understand how you can identify the ongoing replicated artifacts, by simply checking the DB.  Could you add more details?   

I'll also double check the code to refresh my memory about details replication

proposals/new/single-active-replication.md
```diff
@@ -0,0 +1,106 @@
...
+A new option, **"Single Active Replication"**, will be added in the replication policy UI to ensure that replication jobs for the same artifact do not run simultaneously. The default state will be **unchecked**, meaning replication tasks can still run in parallel unless the user opts for single execution.
+
+When the "Single Active Replication" option is enabled, any replication task for the same artifact will not start until the current replication for that artifact finishes. This ensures that bandwidth is not overloaded and the queues are better managed.
+
+Additionally, the implementation will involve adding a **single_active_replication** column in the replication policy in db and updating the worker execution logic to skip replication if a task is already running.
+
+## Changes Made
+
+- Added a **"Single active replication"** checkbox in the replication policy UI.
+- Implemented a best-effort check to avoid concurrent executions of the same replication policy by inspecting ongoing replication tasks.
```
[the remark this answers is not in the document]
bupd: Sorry, It was wrong wording.

will add the out of scope. and update the wording

proposals/new/single-active-replication.md
```diff
@@ -0,0 +1,119 @@
...
+
+A **"Single Active Replication"** checkbox will be added in the replication policy UI. By default, it will be unchecked.
+
+![image](https://github.com/user-attachments/assets/a6d10236-577b-4249-9763-7b8584c2a426)
+
+![Screenshot_2025-01-08_18-44-27](https://github.com/user-attachments/assets/3dcc0d84-68dc-49a1-a51d-b578189cb244)
+
+### Harbor Core
+A new condition is added in the replication flow checking if single_active_replication is enabled for the replication policy. if a replication policy has single_active_replication enabled, the system first checks the database for any currently running executions of that policy.
+- If a execution with running status is found, a new replication job is created but **marked as "skipped"** instead of replication being executed normally.
```
reasonerjt, line 75: By "a new replication job is created", do you mean "a record will be inserted to execution table"?

bupd: yes.

PR #257, opened 2025-01-15
feat: add the model processor proposal

chlins:
Issue: https://github.com/goharbor/harbor/issues/21229


proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+
+![license](../images/ai-model/license.png)
+
+5. Add a new tab named `Files` to the artifact detail page, which will display the files list of the AI model.
+
+![files](../images/ai-model/files.png)
+
+## API
+
+There is no need to introduce new APIs for this proposal, as the existing "addition" API for the artifact can be utilized to extend capabilities for parsing AI model metadata.
```
vivek, line 87: In example below under `Endpoints` it specifies GET calls for `license` or `README.md` or `files`. If there is a new thing added to the model layer `artifactType`, would it continue to work or this new type will need to be added in the `AI Model Processor`. For example:
`/additions/someNewthing`  would continue to work or it would needed to be added in `AI Model Processor`? If it does, then it can be problematic as it will need code change, upgrade of Harbor etc.

Can we leverage `artifactType` to be sent in an HTTP accept header. something like:

```
X-Accept-AI: application/vnd.cnai.model.someNewThing.v1.tar;version=1.3
```
There is already precedence of such header for vulnerabilities `X-Accept-Vulnerabilties`.

`AI Model Processor` can make use of this header along with value in path after `additions` to generically return the layer requested.

vivek (changes requested):
Thanks @chlins for authoring this proposal. Overall it looks good. The key aspects of the proposal make it appealing. My takeaways:

* This proposal brings AI models as first class citizen in Harbor. The user stories to use this CNAI model spec which follows OCI spec to surface AI model
* Uses existing OCI spec (artifactType and mediaType ) along with following CNAI model format  https://github.com/CloudNativeAI/model-spec/blob/main/docs/spec.md
* No additional APIs needed as it leverages additions API - which is good approach

I left a comment regarding future model format changes and how new additions can be handled in generic way without requiring code changes. PTAL, I might be off as new to Harbor at code level, as long as we can handle such use case, its all looking good.

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+## Goals
+
+1. Ability to recognize that the AI model which based on `artifactType` and display the AI icon.
+2. Provide support for parsing AI model metadata based on the [Model Spec](https://github.com/CloudNativeAI/model-spec).
+3. Display the README, LICENSE and Files List of the AI model in the UI. (By existing artifact addition mechanism)
+
+## Non-Goals
+
+1. Ensure the current version is stable and there will be no subsequent changes, and no compatibility issues.
+Because the current model spec is still in a relatively early stage, there may be iterations and updates in the future.
```
reasonerjt, line 27: Can we ensure backward-compatibilities in future updates?   If not we need a way to manage it, for example add a switch and disable the processor by default.

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+    },
+    {
+      "mediaType": "application/vnd.oci.image.layer.v1.tar",
+      "digest": "sha256:44a6e989cc7084ef35aedf1dd7090204ccc928829c51ce79d7d59c346a228333",
+      "size": 5632,
+      "annotations": {
+        "org.cnai.model.filepath": "README.md",
+        "org.cnai.model.readme": "true"
+      },
+      "artifactType": "application/vnd.cnai.model.layer.v1.tar"
```
reasonerjt, line 120: This seems not consistent with the [spec](https://github.com/CloudNativeAI/model-spec/blob/main/docs/spec.md) where it says the artifactType should be `application/vnd.cnai.model.doc.v1.tar`

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+    },
+    {
+      "mediaType": "application/vnd.oci.image.layer.v1.tar",
```
reasonerjt, line 123: This seems not consistent with the [spec](https://github.com/CloudNativeAI/model-spec/blob/main/docs/spec.md) where it says the artifactType should be `application/vnd.cnai.model.config.v1.tar`

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+      "digest": "sha256:a4e7c313c8addcc5f8ac3d87d48a9af7eb89bf8819c869c9fa0cad1026397b0c",
+      "size": 2560,
+      "annotations": {
+        "org.cnai.model.config": "true",
```
reasonerjt, line 127: "org.cnai.model.config" is not a predefined annotation according to 
https://github.com/CloudNativeAI/model-spec/blob/main/docs/annotations.md#layer-annotation-keys

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+    "org.cnai.model.format": "tensorflow",
+    "org.cnai.model.name": "Qwen2.5-0.5B",
+    "org.cnai.model.param.size": "0.49B",
+    "org.cnai.model.precision": "int8",
+    "org.cnai.model.quantization": "gptq"
+  }
+}
+```
+
+### Endpoints
```
reasonerjt, line 196: I have some questions regarding the implementation of these APIs.

In the [spec](https://github.com/CloudNativeAI/model-spec/blob/main/docs/spec.md)  it mentions the annotations are optional.  So we should also clarify what to do with a model that does not provide the annotations.

Esp. for the filelist API, the spec doesn't requirement a 1:1 mapping between a file and a layer, and there can be cases when the annotation `org.cnai.model.filepath` is not added to any or every layer, shall we also consider that?

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+## References
+
+1. [Model Spec](https://github.com/CloudNativeAI/model-spec) (Cloud Native Artifacial Intelegence Model Format Specification)
+2. [modctl](https://github.com/CloudNativeAI/modctl) (Command-line tools for managing OCI model artifacts which are bundled based on Model Spec)
+
+## To Be Discussed
+
+Should we think about offering the capability to open files? like `config.json` and `tokenizer.json`, but how should we handle model files? One approach could be to not display them or only show certain parameter information in the header, similar to what HuggingFace does.
```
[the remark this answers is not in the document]
chlins: Sure, the "open" capability needs to do more design and API changes, btw, the "downloading" may also need the API change as it require backend return the stream for frontend. So I suggest let's put "open" & "downloading" in the phase2.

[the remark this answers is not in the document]
chlins: This part needs to more detailed investigation and design, the model always be large, so I am not sure whether it should return the full model content.

proposals/new/AI-model-processor.md
chlins: A more general mechanism may require a comprehensive restructuring of this part, currently each processor needs to implement the corresponding interface, and for different additions, corresponding code needs to be written for parsing, so a generic way may not be within the scope of this proposal.

proposals/new/AI-model-processor.md
chlins: Updated.

proposals/new/AI-model-processor.md
chlins: Updated.

proposals/new/AI-model-processor.md
chlins: Updated.

proposals/new/AI-model-processor.md
chlins: In the current version, I think we may not need to consider compatibility issues for the time being, but starting from some future stable version, backward compatibility should be guaranteed.

proposals/new/AI-model-processor.md
chlins: Since this value is in the annotation, it might not be suitable as a MUST. Therefore, the handling in Harbor can be considered simpler: for layers with this annotation, return them; for those without, ignore them.

reasonerjt: By "ignore" do you mean the `GET README` or `GET LICENSE` API will return 404 if there is no corresponding `org.cnai.model.filepath` annotation added?  I'm fine with that but could you clarify it in the design?
And as for `GET Files List` API,  do you mean it will just return whatever it knows based on the annotation?  if there're not any `org.cnai.model.filepath` annotations will it return a 404 or an empty list?  It will make this API quite unreliable...

proposals/new/AI-model-processor.md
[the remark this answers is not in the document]
reasonerjt: OK sounds good to me

proposals/new/AI-model-processor.md
reasonerjt: Per discussion, we clarified that we don't expect to have break change in the API, I'm resolving this one.

proposals/new/AI-model-processor.md
steven-zou: The overall model specification relies heavily on specific annotations such as "org.cnai.model.filepath". The proposal is how to support the model specification in the harbor, so is it possible to support validation of the model specification? An extreme example is that if "org.cnai.model.filepath "is not set for each layer file, the processor can't read any file information, and the actual file situation becomes difficult to understand (In fact, it is usable, but from the API to see if there is anything).

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Motivation
+
+Effective model management is increasingly crucial due to the growing popularity of AI/ML applications.
+While Harbor currently focuses on container image management with features like vulnerability scanning, replication, signature and more.
+Harbor is also natively an OCI Artifact registry. Therefore, packaging AI models as OCI Artifacts and storing them in Harbor is a viable approach.
+This would allow leveraging Harbor's existing features for improved AI model management, enhancing its applicability in AI/ML workloads and promoting model reusability and traceability.
+
+
+## Goals
```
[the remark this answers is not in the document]
s3rj1k: https://github.com/CloudNativeAI/modctl/issues/50

s3rj1k:
Just for reference, I am using below `Modelfile` (https://huggingface.co/google/gemma-2-2b), `Modelfile` is inside artifact directory (gemma-2-2b/Modelfile)
```
# https://huggingface.co/google/gemma-2-2b/tree/main

NAME gemma-2-2b
ARCH transformer
FAMILY gemma
FORMAT safetensors
PARAMSIZE 2
PRECISION bf16
QUANTIZATION none

CONFIG config.json
CONFIG generation_config.json
CONFIG special_tokens_map.json
CONFIG tokenizer_config.json
CONFIG tokenizer.json
CONFIG model.safetensors.index.json

MODEL tokenizer.model
MODEL *.safetensors
```

Below how `oras` sees above model after it is pushed into registry

```
# oras manifest fetch domain.net/library/gemma-2-2b:latest | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.cnai.model.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.cnai.model.config.v1+json",
    "digest": "sha256:c9f3f1a5131634d8a8c59f2085bf6745a6b2f1f13bd80019d5635f74b57e2372",
    "size": 264
  },
  "layers": [
    {
      "mediaType": "application/vnd.cnai.model.doc.v1.tar",
      "digest": "sha256:f2bd54bd4695332a5c9fffb5bd736aa3a47a933c03bd4fa1fca508c9f607a1a1",
      "size": 27648,
      "annotations": {
        "org.cnai.model.filepath": "README.md"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:0ad46a0e2c2bfb03aeacef30f5cfdf08fc57e1e2109902594a955d1501d6f998",
      "size": 2048,
      "annotations": {
        "org.cnai.model.filepath": "generation_config.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:4c09b0c73a63882fbbd689a085ec52e83f42f97a2c07ffaaeecdab14ca922aaf",
      "size": 2560,
      "annotations": {
        "org.cnai.model.filepath": "special_tokens_map.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:821088b4b19de1ee9564acc8983345a68150839da75755f50fbb78c8013ad0ea",
      "size": 48128,
      "annotations": {
        "org.cnai.model.filepath": "tokenizer_config.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:3e86b2521cd1e5516a4c7aa09efeb94af5cfb6d842bcb6f4b84ff798ede8a4ea",
      "size": 17527296,
      "annotations": {
        "org.cnai.model.filepath": "tokenizer.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:d85985abcd1e05389abb9088251ca703c3307f946aa693d549f0967558f709f4",
      "size": 26112,
      "annotations": {
        "org.cnai.model.filepath": "model.safetensors.index.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.config.v1.tar",
      "digest": "sha256:dbc9cf744f8ff3e596924765d8a91c7bed4b7939c73b18a373905b1b31d81b85",
      "size": 2560,
      "annotations": {
        "org.cnai.model.filepath": "config.json"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.v1.tar",
      "digest": "sha256:746911d5850fdd7bc058de4f294c02d282f4e99422df18d31dfb83beb7aa8eed",
      "size": 4242944,
      "annotations": {
        "org.cnai.model.filepath": "tokenizer.model"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.v1.tar",
      "digest": "sha256:2d1be53f694599fe05d2f57ba6b9028e766f8a585225d87643ad984c351dd7b3",
      "size": 4992578048,
      "annotations": {
        "org.cnai.model.filepath": "model-00001-of-00003.safetensors"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.v1.tar",
      "digest": "sha256:5fa6ff435737b221e75b771450c9c09b0a3695c73815776450adab71b24ab4a1",
      "size": 4983444992,
      "annotations": {
        "org.cnai.model.filepath": "model-00002-of-00003.safetensors"
      }
    },
    {
      "mediaType": "application/vnd.cnai.model.weight.v1.tar",
      "digest": "sha256:703f96fc5ba4fee31dd14b537da139da015bb4d3351a212209050c2e68acfc75",
      "size": 481383424,
      "annotations": {
        "org.cnai.model.filepath": "model-00003-of-00003.safetensors"
      }
    }
  ]
}
```

as opposed to `Kitfile` from KitOps 

```
# oras manifest fetch domain.net/library/gemma-2-2b:latest | jq
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.kitops.modelkit.config.v1+json",
    "digest": "sha256:4b58400f17b967152ab8ac022946896cd55575f6076e2f79781f1e6e21ffbd0a",
    "size": 2170
  },
  "layers": [
    {
      "mediaType": "application/vnd.kitops.modelkit.model.v1.tar",
      "digest": "sha256:f3b5a4b1c6065dbdeac2afa977d60b110bb84a10455f6e0d6c94c33b7a97cdeb",
      "size": 4992578048
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:b22bcdad944d6588dd92314753482cc2c2265e5231425e2846529a51a9b2393a",
      "size": 4983444992
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:d8554286602a841a31bd9a88d3fa2ec85d6aa8b1eb9b585efb10681d22beae35",
      "size": 481383424
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:d41f170f157a76a8e6fe223726bace7d1b3eb865532488297d96bb5a37adfdcc",
      "size": 2560
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:d77ba08b9e8b7c7cd4dfe49f0df71f17d7f9173280ea23b3e8ad35e4a2eef1bd",
      "size": 2048
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:fa1f3ea472adaaf49d62295483bc7423cccdc62e67064b0acc3a2103866a5057",
      "size": 26112
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:2428f162d376dd61b077463d15ba0a04300697ba888686515d93afdd6fdeb403",
      "size": 2560
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:5d6bc03de874683603f9bae07c7c79195cad9821e3ed5484361ec59c7191e5ae",
      "size": 17527296
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.modelpart.v1.tar",
      "digest": "sha256:d922ba9526e31d91f7d757c4634ab5d861ed433c19a9906faf7715825a025713",
      "size": 48128
    },
    {
      "mediaType": "application/vnd.kitops.modelkit.code.v1.tar",
      "digest": "sha256:79495fe9531fb5f2eb8181a64ffe6659613f10254d1ad7bc5ced38f098550453",
      "size": 4245504
    }
  ],
  "annotations": {
    "ml.kitops.modelkit.cli-version": "1.1.0"
  }
}
```

proposals/new/AI-model-processor.md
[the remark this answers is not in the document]
chlins: We don't want to be bound or limited to a specific tool. As a server-side component, Harbor doesn't need to be concerned with the client-side implementation. Therefore, in theory, any tool based on the model spec implementation can be used to build model artifacts and push them to Harbor for management. `https://github.com/CloudNativeAI/modctl` is one option.

proposals/new/AI-model-processor.md
chlins: Added clarification in the endpoints.

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,323 @@
...
+
+* User Stories (Model-Specific)
+
+1. As a user with read permission of the model artifact, I can distinguish and view the AI model icon besides the artifact digest in the UI.
+2. As a user with read permission of the model artifact, I can view the AI model metadata such as model architecture, name, family, format, etc.
+3. As a user with read permission of the model artifact, I can click the `Readme` tab to view the README of the AI model.
+4. As a user with read permission of the model artifact, I can click the `License` tab to view the LICENSE of the AI model.
+5. As a user with read permission of the model artifact, I can click the `Files` tab to view the files list of the AI model.
+
+* User Stories (Common-Case for OCI Artifact)
```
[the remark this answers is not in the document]
chlins: Updated.

reasonerjt (approved):
lgtm

vivek (approved):
lgtm

proposals/new/AI-model-processor.md
```diff
@@ -0,0 +1,350 @@
...
+
+Response Explanation:
+
+| Status Code | Description |
+| --- | --- |
+| 200 | normal case |
+| 404 | layer with README characteristics not found |
+| 500 | internal server error(other runtime error or unexpected error) |
+
+`GET /api/v2.0/projects/{project_name}/repositories/{repository_name}/artifacts/{reference}/additions/readme`
```
s3rj1k, line 193: should endpoint end in `readme.md`, like for other processors?

s3rj1k:
We also should consider changing model name, reference here https://github.com/goharbor/harbor/pull/21663/files#r1967296025

"MODEL" -> "CNAI" for ArtifactTypeCNAI

Harbor can have multiple ML model processors for different specs, having this "MODEL" is so general, it will look strange when other models specs processors will be added.

chlins:
> We also should consider changing model name, reference here https://github.com/goharbor/harbor/pull/21663/files#r1967296025
> 
> "MODEL" -> "CNAI" for ArtifactTypeCNAI
> 
> Harbor can have multiple ML model processors for different specs, having this "MODEL" is so general, it will look strange when other models specs processors will be added.

@s3rj1k Sure, we can change this name to "CNAI" for future extension.

stonezdj (approved):

bupd (approved):
lgtm

wy65701436 (approved):
lgtm

PR #259, opened 2025-02-19
Add the design for the API to help rotate secret.

reasonerjt:
It decrypt the data with the old key and encrypt using the new key


stonezdj (approved):

Vad1mo (approved):
good thing to have

PR #261, opened 2025-04-16
Add proxy-cache-registry-head-calls.md

mxm-tr:


thcdrt (approved):
LGTM

proposals/new/proxy-cache-registry-head-calls.md
```diff
@@ -0,0 +1,57 @@
...
+1. Use a new cache key in the [proxy controller cache](https://github.com/goharbor/harbor/blob/bfc29904f96e17248a4e6204d12058c1d7d05ab8/src/controller/proxy/controller.go#L78), such as:
+
+```
+cache:manifestexists:<repo>:<ref>
+```
+
+2. Define its lifetime to a value that would prevent rate limiting from being triggered (10s?) in the [proxy-controller](https://github.com/goharbor/harbor/blob/bfc29904f96e17248a4e6204d12058c1d7d05ab8/src/controller/proxy/controller.go#L41-L48)
+
+```golang
+manifestExistsCacheInterval = 10 * time.Second
```
renmaosheng, line 48: Hi Maxime, by given such internal, there would be a a chance the client will put different artifacts with the same binary from the source repo, if your pipeline is using this, you will have different testing results depending on when you run the tests.  Such ambiguity will lead to difficulty when you trouble-shooting some issues.  In my opinion, we should encourage the user to use sha digest instead of tag to reduce the API calling number.

renmaosheng: Hi Maxime, I don't know how to make it configurable, but we should highlight to the users about the possibility of pulling different binaries in different time, thanks.

PR #265, opened 2025-07-21
Garbage Collection Performance Enhancement -- avoid landing tag files in storage

wy65701436:


proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+    On artifact pull/push, Harbor proxies the request with digest instead of tag.
+
+    No more tag deletion API calls to backend are needed during GC, the tag deletion becomes instant (DB-level operation only).
+
+    GC becomes faster, especially in object storage.
+
+## Non Goals
+
+    This proposal does not attempt to clean up previously written tag link files.
```
Vad1mo, line 68: Will it be possible to disable the feature?

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+This proposal presents three options for improving GC performance, with Option 3 as the preferred and proposed solution.
+
+### Option 1: Skip Tag Deletion
+
+    Introduce a user-configurable option to skip tag deletion in the backend when removing manifests. This would eliminate the API call that slows down GC but leaves orphaned tag files in the backend.
+
+#### Pros: Simple to implement, immediate performance gain.
+#### Cons: Leaves tag files behind, may cause confusion or inconsistencies for users browsing storage directly.
+
+### Option 2: Batch Tag Deletion via Upstream Patch
```
Vad1mo, line 28: what effort do you expect in implementing option #3, instead of switching to latest upstream registry, as proposed in #2?

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+    Backport the upstream change from Distribution to support batch deletion of tags (introduced in later versions of the registry). This allows multiple tag links to be deleted in one API call.
+
+#### Pros: Compatible with current Distribution behavior, improves performance.
+#### Cons: Limited gain, still depends on Distribution API and backend performance.
+
+### Option 3: Do Not Land Tag Files in Backend (Proposed)
```
Vad1mo, line 35: Is this implementation not just a stop gap measure and will be removed once distribution v3 is used in Harbor?

Vad1mo (changes requested):

proposals/new/gc_perf_enhancement.md
wy65701436: no, the proposal is a permanent since the tag link is no longer stored in the storage.

wy65701436:
> I think options 2 and 3 can work together to complement each other. For historical stock images, option 2 can improve performance. For newly built image scenarios, option 3 ensures that there are no tags in the distribution. In this case, calling its tag deletion API should theoretically also be fast (perhaps this needs to be confirmed?). This way, we can address the issue of slow garbage collection for both existing and newly built images.

I don't understand clearly, the option 3 is to remove the tag deletion permanently. How can we make those two options work together?

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+    GC becomes faster, especially in object storage.
+
+## Non Goals
+
+    This proposal does not attempt to clean up previously written tag link files.
+
+## Compatibility and Consistency
+
+    No breaking changes; tag permissions are enforced at the API level.
```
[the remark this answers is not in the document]
Vad1mo: Yes, from the outside OCI interfaces, yes, but internally you swap parts of the logic implemented in distribution with a different logic.

Vad1mo (changes requested):
Slow GC performance , has been a problem in Harbor and Distribution since day one, so for more than 10 years. 

- This issue affects <0.1%-1% of users; >99% will never run into this issue.
- It only impacts users who are using object store (afaik more on Azure than S3).
- It only impacts users who are running GC occasionally. (every couple of months). 
- The issue won't occur if the GC runs eg. weekly/monthly.

## Option 3: One-Way Door Decision, Split Logic, No Way Back
I think Option no. 3 is the worst option for the future of Harbor. Because this is a [one-way door decision](https://www.youtube.com/watch?v=rxsdOQa_QkM), although even two two-way-door decisions exist. 
- **No option to disable the feature** 
- **Not futureproof**
- Solution for ~0.1% of users while having an unpredictable impact on 100% of users.
- If a better solution is introduced in distribution in the future, Harbor can only adopt it with a lot of effort.
- Extremely difficult to debug and trace down errors in operation correctly. E.g what happens if transaction rollbacks?
- Split logic, as some data is in the DB and some data is in the object store. There is a high risk that things can get out of sync.
- Creating a backup of only images becomes impossible. 
- Image/object store and DB need to be in sync; this is not the case yet, but currently this only affects some rare edge cases.
- If DB and object store are out of sync, what is the strategy to fix it? 

## Option 2: Take Patch From Upstream Distribution V3 or Tell Users to Use V3 
- **While Harbor currently only supports V2, there is nothing stopping users from using V3 instead of V2.** 
- This is a fix in upstream distribution v3 -  https://github.com/distribution/distribution/pull/4329
- It does not look problematic to backport it, since we already forked V2
- When Distribution V3, is adopted in Harbor the problem will be solved upstream.
- Forward and backward compatible.  
- Users are already adopting V3 with Harbor.

## Option 1: Allow Skip Tag Deletion if Users Really Need It
As said, this is the simplest option.
- **Will provide relief to 0.1% of users suffering from this issue while not impacting the other 99%**
- **Can be enabled or disabled if needed**
- Future proof
- Forward and backward compatible
- Tags can be deleted out of bounds with a script if needed


**Going forward, I am suggesting adapting with a feature switch, Option 1 or Option 2.**

bupd (commented):
@wy65701436 Thank you for this proposal. This is something the community has been asking for so long time.

Here are my suggestions
The current idea stated in proposal only solves part of the performance issue. (i.e not storing tags in the filesystem) - it’s like crossing half the river. If Harbor also stores both digests and their tag relations directly in its DB, we can remove the need for digest lookups during garbage collection done by distribution.

From what we have observed, the time taken to traverse cross-linked blobs by the distribution is actually greater than the time taken to delete an item. This becomes even more of a problem in S3-like backends where traversal is far more expensive than put or delete operations.

A better long-term approach is to upgrade to distribution v3. This enables batch deletion of unreferenced blobs which is very helpful for deleting multiple blobs in a single request and pairing this with Harbor GC to run entirely at the DB level, without relying on distribution to traverse references. would be a great perf improvement.

So, distribution will only be responsible for storing and deleting objects, while Harbor manages all reference tracking. This removes costly traversal, simplifies GC overall.

wy65701436:
> @wy65701436 Thank you for this proposal. This is something the community has been asking for so long time.
> 
> Here are my suggestions The current idea stated in proposal only solves part of the performance issue. (i.e not storing tags in the filesystem) - it’s like crossing half the river. If Harbor also stores both digests and their tag relations directly in its DB, we can remove the need for digest lookups during garbage collection done by distribution.
> 
> From what we have observed, the time taken to traverse cross-linked blobs by the distribution is actually greater than the time taken to delete an item. This becomes even more of a problem in S3-like backends where traversal is far more expensive than put or delete operations.
> 
> A better long-term approach is to upgrade to distribution v3. This enables batch deletion of unreferenced blobs which is very helpful for deleting multiple blobs in a single request and pairing this with Harbor GC to run entirely at the DB level, without relying on distribution to traverse references. would be a great perf improvement.
> 
> So, distribution will only be responsible for storing and deleting objects, while Harbor manages all reference tracking. This removes costly traversal, simplifies GC overall.

Thanks @bupd for your comments.

I’m not opposed to upgrading to v3 from Harbor’s perspective—in fact, I support it since I’m also one of its maintainers. However, one concern we’ve already discussed in the community is that we’d like to wait until v3 is stable before making the move, as many Harbor users run it in production environments.

That said, upgrading to v3 is not within the scope of this proposal. Option 3 does not prevent us from upgrading in the future. I understand the benefits of v3, but this is a separate topic. Even if we move to v3, as long as Harbor continues to store tag files in the backend, we will still need to handle tag deletion regardless of any performance improvements.

For the reasoning behind my proposal to avoid storing tags in the backend, please see the earlier comments.

In summary, this proposal is intended to drive discussion based on the current situation, which is still built on distribution v2.

bupd:
@wy65701436 I understand your concern about v3 not being fully mature, but I’m not too worried about v3 at this stage. What I’d really like your thoughts on the traverse problem.
> When running GC the time taken to traverse cross-linked blobs by the distribution is actually greater than the time taken to delete an item. This becomes even more of a problem in S3-like backends where traversal is far more expensive than put or delete operations.

Without addressing the traversal overhead, This would be like crossing half the river. If we do fix this traversing problem we will see huge improvements in GC performance. (i.e days to ~hours) improvement. The larger the registry the slower the traversal gets for unreferenced blobs.

wy65701436:
> @wy65701436 I understand your concern about v3 not being fully mature, but I’m not too worried about v3 at this stage. What I’d really like your thoughts on the traverse problem.
> 
> > When running GC the time taken to traverse cross-linked blobs by the distribution is actually greater than the time taken to delete an item. This becomes even more of a problem in S3-like backends where traversal is far more expensive than put or delete operations.
> 
> Without addressing the traversal overhead, This would be like crossing half the river. If we do fix this traversing problem we will see huge improvements in GC performance. (i.e days to ~hours) improvement. The larger the registry the slower the traversal gets for unreferenced blobs.

@bupd 

I believe there are two main areas we can improve:

- Tag deletion performance in v2 – this needs to be addressed regardless of which option we choose, even if none of the three proposed options are ideal.
- Performance benefits by moving to v3 – upgrading to v3 could bring additional improvements.

These two areas are independent and can be tackled separately without conflict. For this proposal, the focus is on area-1. For area-2, we can start a separate discussion thread.

And I would like to hear your suggestions on the three options, and especially the option 3 that I proposed.

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Pros: Compatible with current Distribution behavior, improves performance.
+#### Cons: Limited gain, still depends on Distribution API and backend performance.
+
+### Option 3: Do Not Land Tag Files in Backend (Proposed)
+
+    Modify Harbor's proxy logic to avoid writing tag link files to the backend. When a manifest with a tag is pushed:
+       
+        Harbor Core will extract the tag and persist it in its own database.
+        The proxy request to the backend registry is rewritten to use the manifest's digest instead of the tag (PUT /v2/<repo>/manifests/<digest>).
```
bupd, line 40: and a bit more technically elaborate on what is current harbor flow and what is the harbor flow in future in regard to this proposal.

from the perspective of user would be beneficial

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+    Backport the upstream change from Distribution to support batch deletion of tags (introduced in later versions of the registry). This allows multiple tag links to be deleted in one API call.
+
+#### Pros: Compatible with current Distribution behavior, improves performance.
+#### Cons: Limited gain, still depends on Distribution API and backend performance.
+
+### Option 3: Do Not Land Tag Files in Backend (Proposed)
+
+    Modify Harbor's proxy logic to avoid writing tag link files to the backend. When a manifest with a tag is pushed:
+       
+        Harbor Core will extract the tag and persist it in its own database.
```
bupd, line 39: please add that harbor is already doing this.

wy65701436:
> @wy65701436 I understand your concern about v3 not being fully mature, but I’m not too worried about v3 at this stage. What I’d really like your thoughts on the traverse problem.
> 
> > When running GC the time taken to traverse cross-linked blobs by the distribution is actually greater than the time taken to delete an item. This becomes even more of a problem in S3-like backends where traversal is far more expensive than put or delete operations.
> 
> Without addressing the traversal overhead, This would be like crossing half the river. If we do fix this traversing problem we will see huge improvements in GC performance. (i.e days to ~hours) improvement. The larger the registry the slower the traversal gets for unreferenced blobs.

As for the traversing, Harbor doesn't rely on the distribution to locate the file but the harbor DB. So, we don't have any traversing performance problem.

proposals/new/gc_perf_enhancement.md
wy65701436: Discussed with @bupd offline, I've already added in the proposal, mark it as resolved.

bupd (approved):
lgtm

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+Harbor currently proxies manifest and tag pushes to the backend registry (Distribution). This results in both manifest blobs and tag link files being written to the underlying storage. 
+During garbage collection, each tag link must be removed explicitly via a Distribution API call, which is slow in object storage setups. For example, S3 may take multiple seconds per tag deletion.
+
+## Solution
+
+This proposal presents three options for improving GC performance, with Option 3 as the preferred and proposed solution.
+
+### Option 1: Skip Tag Deletion
+
+    Introduce a user-configurable option to skip tag deletion in the backend when removing manifests. This would eliminate the API call that slows down GC but leaves orphaned tag files in the backend.
```
reasonerjt, line 23: I think this "user-configurable option" will be an option for GC job right?  Could you clarify.

Please fix the format, for some reason it looks like quoted content.  It should be normal test which looks like section "Solution"

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Pros: Simple to implement, immediate performance gain.
+#### Cons: Leaves tag files behind, may cause confusion or inconsistencies for users browsing storage directly.
+
+### Option 2: Batch Tag Deletion via Upstream Patch
+
+    Backport the upstream change from Distribution to support batch deletion of tags (introduced in later versions of the registry). This allows multiple tag links to be deleted in one API call.
```
reasonerjt, line 30: See comment https://github.com/goharbor/community/pull/265/files#r2347885753

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Option 3: Do Not Land Tag Files in Backend (Proposed)
+
+    Modify Harbor's proxy logic to avoid writing tag link files to the backend. When a manifest with a tag is pushed:
+       
+        Harbor Core will extract the tag and persist it in its own database.
+        The proxy request to the backend registry is rewritten to use the manifest's digest instead of the tag (PUT /v2/<repo>/manifests/<digest>).
+
+#### Pros: Eliminates backend tag file overhead completely, simplifies GC logic, significant performance gain.
+#### Cons: Existing tag files will not be deleted unless explicitly handled.
```
reasonerjt, line 43: I think this "Con" is not negligible b/c with the change proposed in option 3, the existing users do not get any benefit.  The GC will still possibly take very long time.

proposals/new/gc_perf_enhancement.md
reasonerjt, line 39: I think you can add a little background to explain currently there are already some tags are stored in database only.

proposals/new/gc_perf_enhancement.md
stonezdj, line 35: If there is a configure item or env variable, how use enable/disable this feature?

proposals/new/gc_perf_enhancement.md
wy65701436: added

proposals/new/gc_perf_enhancement.md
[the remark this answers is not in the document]
wy65701436: The option is removed, this proposal will not discuss the backport as well as the distribution bump.

proposals/new/gc_perf_enhancement.md
wy65701436: no, but we do have a option to toggle the tag deletion for existing images.

Vad1mo (approved):

chlins (approved):
lgtm

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Pros: Simple to implement, immediate performance gain.
+#### Cons: Leaves tag files behind, may cause confusion or inconsistencies for users browsing storage directly.
+
+### Option 2: Batch Tag Deletion via Upstream Patch
+
+    Backport the upstream change from Distribution to support batch deletion of tags (introduced in later versions of the registry). This allows multiple tag links to be deleted in one API call.
+
+#### Pros: Compatible with current Distribution behavior, improves performance.
+#### Cons: Limited gain, still depends on Distribution API and backend performance.
```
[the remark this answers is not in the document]
wy65701436: The MR I want to backport introduces parallel tag deletion, but it still relies on calling the same API. That’s why I described it as a “limited gain.”

It doesn’t reduce the number of API calls, since the parallel deletion happens within each API call, but it does help shorten the overall execution time.

Anyway, I've removed the option2.

proposals/new/gc_perf_enhancement.md
wy65701436: updated.

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Option 3: Do Not Land Tag Files in Backend (Proposed)
+
+    Modify Harbor's proxy logic to avoid writing tag link files to the backend. When a manifest with a tag is pushed:
```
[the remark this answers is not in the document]
wy65701436: updated.

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,131 @@
...
+    Remove the latest tag from the UI.
+    Attempt to pull harbor-instance/library/hello-world:latest — the client will report not found.
+    However, the tag file still exists in the backend storage at this point because Harbor does not sync it, nor is there a need to.
+
+
+### Tag Landing
+
+Push with Tag (current behavior):
+
+![Data Flow Diagram with Tag](../images/gc-perf/push_with_tag_flow.png)
```
reasonerjt, line 38: I think it would be better to update the diagrams of both "current behavior" and "proposed behavior" to specify the URI of the request that is sent from Client to Harbor Core.

So it's clearer how we changed the request.

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,131 @@
...
+
+Modify Harbor's proxy logic to avoid writing tag link files to the backend. When a manifest with a tag is pushed:
+
+1. Harbor Core will extract the tag and persist it in its own database. 
+2. The proxy request to the backend registry is rewritten to use the manifest's digest instead of the tag (PUT /v2/<repo>/manifests/<digest>).
```
reasonerjt, line 43: Please check the .md file, you should not use the "<" sign b/c it has meaning in html.

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,131 @@
...
+
+Push with Digest (Proposed Behavior):
+
+![Data Flow Diagram with Digest](../images/gc-perf/push_with_digest_flow.png)
+
+We would see the Tag is not landed in the background storage.
```
reasonerjt, line 49: I'm still thinking MAYBE we should provide an option to user just in case they WANT the tags to be landed to storage for whatever reasons, easier to switch to other registry?

If you think it's better not to provide this option, could you explicitly explain the rationale in this design?

proposals/new/gc_perf_enhancement.md
```diff
@@ -0,0 +1,131 @@
...
+As mentioned earlier, Harbor will no longer create tag files in the backend storage. As a result, tag deletion will not be necessary for the garbage collection process for newly pushed artifacts. The remaining question is how to handle existing tag files during garbage collection.
+
+This proposal introduces a checkbox in the garbage collection configuration to control this behavior:
+
+![Data Flow Diagram with Tag](../images/gc-perf/gc_tag_del_checkbox.png)
+
+1. Checked – Garbage collection will follow the current behavior and delete tag files.
+2. Unchecked – Garbage collection will skip tag file deletion, improving performance.
+
+This option is intended as a temporary measure and will be removed in a future release. By then, most artifacts in Harbor will be tagless, eliminating the need for explicit user configuration.
```
reasonerjt, line 62: I don't think we can remove the flag.  For legacy users when they have tag landed in storage and they choose skip tag deletion in GC.  We should always give them the option to clean up these tags in GC.

stonezdj (approved):

reasonerjt (approved):

proposals/new/gc_perf_enhancement.md
jsingleton785: Hey there, my team is exactly the user that @reasonerjt was advocating for and now have to find a workaround for Harbor not populating the tags to the S3 backend. 

If there was a way to add a switch, we would gladly accept the GC overhead.

moh2a: hello @jsingleton785 , did you find a workaround ?

PR #280, opened 2026-04-02
Add proxy cache filter proposal

stonezdj:


proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,153 @@
...
+4. Provide API and UI support for creating and updating filter configuration as part of proxy cache project configuration.
+
+1. This proposal does not add denylist precedence rules. Initial scope is allowlist-only.
+2. This proposal does not introduce global/system-level proxy filter policy.
+3. This proposal does not support tag-based filtering in this phase, because pull requests by tag are often converted to pull requests by digest in client side.
+4. This proposal does not perform content trust or vulnerability policy evaluation as part of filter matching.
+
+## Solution
+
+Add a new optional setting for proxy cache projects: `repository_filter`.
```
chlins: Is this setting applied to the project metadata? Do we have a universal proxy cache related config section? It maybe confused to other normal projects.

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,153 @@
...
+
+Matching semantics:
+
+1. If `repository_filter` is empty or not set, all repositories are allowed (same as today).
+2. If `repository_filter` is configured, Harbor allows the pull only when the requested repository matches the configured pattern.
+3. The filter format is JSON: `{"filter":"<pattern>","kind":"regex"|"doublestar"}`.
```
chlins: Is this filter only support one pattern as it is not the array?

proposals/new/filter_repository_proxycache.md
stonezdj: It is in project_metadata, means it is configured per proxy cache project.

proposals/new/filter_repository_proxycache.md
stonezdj: Yes, one pattern, because array can be transformed into one pattern.

chlins (approved):
lgtm

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
+Proposal: Limit proxy cache repositories by filter
```
wy65701436, line 1: Just a general question, do we plan to support regex for proxy cache? if so, should we extend this capability to other related functionalities, such as replication and tag retention?

I proposal in v2.16, we keep the scope of double star. And we can introduce an enhancement in v2.17 to provide an new filter kind - regex.

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+## Database Schema Changes
+
+No new table is required for initial implementation when storing as project metadata.
+
+
+## Compatibility
+
+1. Existing proxy cache projects without filters keep current behavior (allow all).
+2. New behavior is opt-in by configuring `proxy_filter_pattern` and `proxy_filter_kind`.
+3. Clients receive `404 Not Found` for repositories that do not match the configured filter.
```
wy65701436: Returning 404 Not Found is a good compatibility/security choice because it avoids revealing whether the repository exists upstream.

But the log should clearly say:

```
proxy cache repository blocked by filter
project=<project>
repository=<repo>
filter_kind=<kind>
filter_pattern=<pattern>
```

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+1. Invalid pattern in `proxy_filter_pattern` is treated as non-match at runtime.
+2. In project metadata API validation, an unsupported `proxy_filter_kind` value is rejected with `400 Bad Request`.
+
+### Data model
+
+Store filter as project metadata, consistent with existing proxy project settings. Both values are plain strings, matching the convention of all existing Harbor metadata keys.
+
+Proposed metadata keys:
+
+- `proxy_filter_pattern`: the filter pattern string (e.g. `library/**` or `^library/.*`).
```
wy65701436:
```
proxy_cache_filter_pattern
proxy_cache_filter_kind
```

more explicit names may be better

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+
+### API changes
+
+Extend project APIs for proxy cache configuration:
+
+1. `POST /api/v2.0/projects`
+2. `PUT /api/v2.0/projects/{project_name_or_id}`
+3. `GET /api/v2.0/projects/{project_name_or_id}` and list APIs should include configured filters in response.
+
+`proxy_filter_pattern` and `proxy_filter_kind` are metadata for proxy cache projects only. For non-proxy projects, these keys are ignored on create/update and omitted from response.
```
wy65701436: why not reject these fields for non-proxy projects with 400 Bad Request?

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+1. Resolve current project and verify it is a proxy cache project.
+2. Load project proxy filter configuration.
+3. Build artifact request context (repository).
+4. Evaluate against repository filter.
+5. If not matched, return `404 Not Found`.
+6. If matched, continue existing proxy logic unchanged.
+
+Notes on behavior:
+
+1. Invalid pattern in `proxy_filter_pattern` is treated as non-match at runtime.
```
wy65701436: Treating invalid patterns as runtime non-matches makes configuration errors hard to detect and can cause unintended silent blockages. Instead, we should compile and validate the pattern during project metadata mutation (POST/PUT APIs) and return 400 Bad Request if the pattern is invalid.

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+Pattern format:
+
+- `regex`: regular expression matching against repository path.
+- `doublestar`: glob-style matching with doublestar support (`*`, `**`, `{a,b}`, and related glob features).
+
+## Implementation Details
+
+### Request enforcement point
+
+Enforcement should happen in proxy middleware before upstream manifest proxy operations:
```
wy65701436, line 59: To ensure a complete fail-closed implementation, we need to explicitly cover all manifest entry points in the middleware pre-check layer. This includes manifest GET and HEAD requests, resolved by either tag or digest. Harbor must return a 404

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+4. This proposal does not perform content trust or vulnerability policy evaluation as part of filter matching.
+
+## Solution
+
+Add two new optional metadata settings for proxy cache projects: `proxy_filter_pattern` and `proxy_filter_kind`.
+
+Matching semantics:
+
+1. If `proxy_filter_pattern` is empty or not set, all repositories are allowed (same as today).
+2. If `proxy_filter_pattern` is configured, Harbor allows the pull only when the requested repository matches the configured pattern.
```
wy65701436: Which repository namespace does the proxy_filter_pattern match against—the upstream registry path (library/redis) or Harbor's local proxy project namespace (dockerhub-proxy/library/redis)?

proposals/new/filter_repository_proxycache.md
Vad1mo: Supporting only double star ** and add regex later makes ** a legacy feature from the beginning, that needs to be supported forever and maintained alongside regex. 

Regexes are a more powerful capability, we should opt-in for doing regex only.

proposals/new/images/filter_repository_proxycache_ui.png
Vad1mo, line 1: Conceptually and logically, project creation and proxy policy should be separated. Like we have a replication policy we should have a proxy policy.  This is a good opportunity to split it. 

On top it would also allow us to have multiple rules per proxy policy.

Vad1mo (commented):
just like other policies in harbor the proxy policy should be first class citizen of Harbor and not a toggle during project creation.

proposals/new/filter_repository_proxycache.md
Vad1mo: Regex anchoring is inconsistent. Line 19 uses unanchored (library|goharbor|myorg)/.*; line 85 uses anchored ^library/.*.


Either auto-anchor (^…$) or document that the author must anchor. **Strongly recommend auto-anchoring**

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+  "proxy_filter_pattern": "library/**",
+  "proxy_filter_kind": "doublestar"
+}
+```
+
+Validation:
+
+1. `proxy_filter_pattern` can be empty (empty means match all).
+2. `proxy_filter_kind` must be `regex` or `doublestar` when specified; defaults to `doublestar` if omitted.
+3. An unrecognised `proxy_filter_kind` value is rejected with `400 Bad Request` during project metadata validation.
```
Vad1mo: this is a contradiction to line 76
> Invalid pattern in `proxy_filter_pattern` is treated as non-match at runtime.

proposals/new/images/filter_repository_proxycache_ui.png
Vad1mo: Image

proposals/new/filter_repository_proxycache.md
bupd:
```suggestion
2. If `proxy_filter_pattern` is configured, Harbor allows new upstream proxying only when the normalized upstream repository path matches the configured pattern. 
```

proposals/new/filter_repository_proxycache.md
```diff
@@ -0,0 +1,156 @@
...
+## Solution
+
+Add two new optional metadata settings for proxy cache projects: `proxy_filter_pattern` and `proxy_filter_kind`.
+
+Matching semantics:
+
+1. If `proxy_filter_pattern` is empty or not set, all repositories are allowed (same as today).
+2. If `proxy_filter_pattern` is configured, Harbor allows the pull only when the requested repository matches the configured pattern.
+3. `proxy_filter_kind` specifies the matching mode: `regex` or `doublestar`. It is optional; if omitted, Harbor uses `doublestar` matching by default.
+4. If the repository does not match, Harbor returns `404 Not Found` and does not request content from upstream.
```
bupd:
```suggestion
4. If the normalized repository does not match, Harbor returns `404 Not Found` for the upstream proxy attempt and does not request content from upstream. Already cached local content is not affected by this filter.
```

I believe things would be better if this proposal only focuses on preventing new cache fills.

proposals/new/filter_repository_proxycache.md
bupd:
```suggestion
3. For existing proxy cache projects where a filter is later configured, the filter only affects new upstream proxy requests and cache fills. Artifacts already cached locally remain accessible through existing harbor proxy behaviour.
4. Clients receive `404 Not Found` for repositories that do not match the configured filter.
```

better more transparent wording.

bupd (changes requested):
I suggest to improve wording. 

@stonezdj

proposals/new/filter_repository_proxycache.md
stonezdj: It follow the previous convention related to proxy cache project, and also provide more robustness to the API if the unused parameter is ignored.

proposals/new/filter_repository_proxycache.md
stonezdj: This feature is used to limit the usage of some sort of in secure image to been proxy cached by Harbor, if we set different rule between upstream content and cached image, it will cause inconsisent and maybe considered as an security bypass.

proposals/new/filter_repository_proxycache.md
stonezdj: Keep the same policy to upstream and local cached images.

proposals/new/images/filter_repository_proxycache_ui.png
stonezdj: Do you have existing implementation for proxy cache filter? If yes, could you please cherry-pick it?
All filter rules can be merged into one loggically, we don't want to implement it so heavy because it is hard to debug complex filter rules. and implement UI and add new table make it more complex.

proposals/new/filter_repository_proxycache.md
stonezdj: Lets follow the previous convention, such as proxy_speed_kb, max_upstream_conn.

proposals/new/filter_repository_proxycache.md
stonezdj: It should be the normalized form `library/redis`

proposals/new/filter_repository_proxycache.md
stonezdj: In current implementation, both doublestar and regular exp are supported, user can select the filter kind, for replication and tag retention, same loggic can be applied, for example
```
[{"type":"name","value":"{firstfloor/hello-world,firstfloor/ldaputils}"},{"type":"tag","value":"**","decoration":"matches"}]
```
It is the replication rule, we can add the kind for each filter like that, the kind can be `doublestar` and `regex`
```
[{"type":"name","value":"{firstfloor/hello-world,firstfloor/ldaputils}", "kind": "doublestar"},{"type":"tag","value":"**","decoration":"matches"}, "kind": "doublestar"]
```
If the kind is empty or not found, then default to doublestar, so there is no need to migrate the legacy data.

proposals/new/filter_repository_proxycache.md
stonezdj: All covered.

proposals/new/filter_repository_proxycache.md
stonezdj: OK, will use auto-anchoring for regex match

proposals/new/filter_repository_proxycache.md
stonezdj: Add a new line:
```
2. Invalid pattern in `proxy_cache_filter_pattern` is rejected with `400 Bad Request` when creating/updating project metadata.
```

bupd (approved):
LGTM

please fix DCO @stonezdj

wy65701436 (approved):
lgtm

